Issue #3, opened 2016-03-16
Performance tests for v6 |  #tf-v6-perf-tests


andreasohlund:
Taskforce: William (Lead) / Igal / Greg / Ramon / Dennis / Hadi @WilliamBZA @hmemcpy @gbiellem @ramonsmits @dvdstelt @HEskandari 

Discussion: [#tf-v6-perf-tests](https://particularsoftware.slack.com/messages/tf-v6-perf-tests/)

We need to run end to end performance tests for v6 transports and persistence for all the supported transport transaction modes. 

I have a hunch that we can reuse the code and infra from the outbox tests?

## Plan of Attack

Action plan can be found [here](https://github.com/Particular/EndToEnd/blob/docs/docs/action-plan.md)

mauroservienti:
@dvdstelt is working on something that is related to this as well

WilliamBZA:
I'd like to participate in this, and I suspect @ramonsmits might like to as well.

WojcikMike:
Count me in

andreasohlund:
Given the potential overlap with #2 how should we play this?

WojcikMike:
@andreasohlund so I assume you prefer to have 2 separate TF ideally without people overlap?

andreasohlund:
Or we start with either #2 or #3 to avoid overlap since both will build on the outbox tests?

WojcikMike:
sounds good to me. Pick one that is more important and one can start with it

andreasohlund:
I'd vote for performance first?

If we agree can you sync with the TF of #2 to decide if they want to join here instead @WojcikMike ?

gbiellem:
@WojcikMike 

>  @gbiellem @johnsimons are you interested and have time please comment on #3

Sure count me in.

ramonsmits:
Would be willing to participate in this as I have created numerous stress tests for SQL, Rabbit and MSMQ.

andreasohlund:
So we have @WojcikMike @gbiellem @ramonsmits @WilliamBZA and potentially @dvdstelt interested in this one. Let's get it going, will take the lead? When can we have a preliminary POA ready?

andreasohlund:
@dannycohen can you chime in with the data want to get out of this from a "this is why you should use v6" perspective?

WilliamBZA:
### Items to discuss and work on

 - [ ] Identify requirement outcomes
   - Do we plan to use the results to lure prospects?
[...]

PR #5, opened 2016-03-22
Moving code base as a starting point for Performance Tests

WojcikMike:
As a initiation of #3 we are moving this codebase


WojcikMike:
@Particular/endtoend-maintainers please merge in, as this is renames solution from: OutboxPerformanceTests

PR #22, opened 2016-04-10
Statistics output to Splunk

hmemcpy:
Writing statistics output to Splunk via NLog network target


src/PerformanceTests/Common/packages.config
```diff
@@ -1,8 +1,8 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <packages>
   <package id="CommandLineParser" version="1.9.71" targetFramework="net45" />
-  <package id="Metrics.NET" version="0.3.4-pre" targetFramework="net452" />
   <package id="Newtonsoft.Json" version="6.0.8" targetFramework="net452" />
+  <package id="NLog" version="3.2.1" targetFramework="net452" />
```
WilliamBZA: Ramon's PR uses NLog v4, can you update?

hmemcpy: I used this version because the Metrics.NET appender didn't work with the newer one.
But since we're not using Metrics.NET at the moment, I can update it.

hmemcpy:
:kiss:

PR #24, opened 2016-04-10
Removing code redundancies according to our style rules

hmemcpy:


ramonsmits:
@hmemcpy This is all based on the Resharper inspections?

hmemcpy:
@ramonsmits yes, this is our shared style from https://github.com/Particular/RepoStandards

PR #26, opened 2016-04-11
Clean & Seed data

dvdstelt:


dvdstelt:
Had a quick talk with @ramonsmits and we got some new ideas I'm going to implement quickly.

dvdstelt:
Can this be merged

src/PerformanceTests/Common.Data/Program.cs
```diff
@@ -32,9 +32,8 @@ static void Main(string[] args)
                 if (Environment.UserInteractive) Console.Title = PermutationParser.ToString(permutation);
 
                 var assembly = System.Reflection.Assembly.GetExecutingAssembly();
-                var tasks = permutation.Tests.Select(x => (IStartAndStop)assembly.CreateInstance(x)).ToArray();
-
-                Run(options, permutation, tasks);
+                var runnableTests = permutation.Tests.Select(x => (BaseRunner) assembly.CreateInstance(x)).ToList();
+                runnableTests.ForEach(s => s.Execute(permutation, options, endpointName));
```
ramonsmits: This will run these tests not in parallel but sequentially.

src/PerformanceTests/Common.Data/Scenarios/BaseRunner.cs
```diff
@@ -0,0 +1,127 @@
...
+        Endpoint.Create(configuration).GetAwaiter().GetResult();
+    }
+
+    IEndpointInstance CreateSendOnlyEndpoint(Configuration configuration)
+    {
+        configuration.SendOnly();
+        return Endpoint.Start(configuration).GetAwaiter().GetResult();
+    }
+
+    IEndpointInstance CreateEndpoint(Permutation permutation, BusCreationOptions options, string endpointName)
```
ramonsmits: Don't pass around those permutation, options and endpointname values. Put the values in fields to that you don't leak all those arguments in all methods.

src/PerformanceTests/Common.Data/Scenarios/ICreateSeedData.cs
```diff
@@ -0,0 +1,22 @@
+﻿namespace Common.Scenarios
+{
+    using NServiceBus;
+
+    interface ICreateSeedData
+    {
+        int SeedSize { get; set; }
```
ramonsmits: Never assigned, probably want to use `ICreateSeedData.SeedSize` ?

src/PerformanceTests/Common.Data/Scenarios/BaseRunner.cs
```diff
@@ -0,0 +1,127 @@
...
+        var endpoint = CreateSendOnlyEndpoint(configuration);
+
+        Parallel.For(0, seedSize, ((i, state) =>
+        {
+            if (i % 5000 == 0)
+                Log.Info($"Seeded {i} messages.");
+
+            ((ICreateSeedData)this).SendMessage(endpoint, endpointName);
+        }));
+        Log.Info($"Seeded total of {seedSize} messages.");
```
ramonsmits: Don't use string interpolotation in logging.

src/PerformanceTests/Common.Data/Program.cs
dvdstelt: You want to run performance tests sequentially? Even if the permutations are completely different, they'd still be running on the same machine.

src/PerformanceTests/Common.Data/Scenarios/ICreateSeedData.cs
dvdstelt: Yep, made mistake in accessing seedsize in incorrect partial class. Fixed it and added Guard that seedsize was properly set.

src/PerformanceTests/Common.Data/Program.cs
ramonsmits: Performance tests should not run sequential nor parallel. Stability runs should have the ability to run multiple scenarios in parallel.

PR #27, opened 2016-04-11
Look at environment variables first for connection strings

WilliamBZA:
First looks for the connection string in the environment variables flag, then falls back to config file

### DIFF
--- a/src/PerformanceTests/Common.Data/IProfileExtensionMethods.cs
+++ b/src/PerformanceTests/Common.Data/IProfileExtensionMethods.cs
@@ -1,0 +1,27 @@
+﻿namespace Common
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Configuration;
+    using System.Text;
+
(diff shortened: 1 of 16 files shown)


src/PerformanceTests/Common.Data/IProfileExtensionMethods.cs
```diff
@@ -0,0 +1,27 @@
...
+                return environmentVariableConnectionString;
+            }
+
+            var applicationConfigConnectionString = ConfigurationManager.ConnectionStrings[connectionStringName].ConnectionString;
+            if (!string.IsNullOrWhiteSpace((applicationConfigConnectionString)))
+            {
+                return applicationConfigConnectionString;
+            }
+
+            return string.Empty;
```
hmemcpy: Should this throw, perhaps? If no connection string is defined, can we continue without?

hmemcpy:
LGTM :+1:

PR #31, opened 2016-04-12
SerializationTests fixes

hmemcpy:
Making Paket iterate all feeds when looking for a matching package


hmemcpy:
@andreasohlund please review.
I decided currently not to include the myget feed, as this needs a bit more work to fit with the Paket API.
I've opened #33 to address this, by replacing Paket API with nuget.

In the meantime, this should pull the latest packages directly from the public nuget feed.

andreasohlund:
Works now for me

PR #43, opened 2016-04-14
Session and category identifiers

HEskandari:
This sends session Id, test category and description to splunk for potential further filtering and aggregation. It also generates the start.bat properly passing all the mentioned values.


HEskandari:
scratch that...this is on a wrong branch...will send a new PR.

PR #44, opened 2016-04-14
Sending category, description and session id to splunk and generating…

HEskandari:
Connects to https://github.com/Particular/EndToEnd/issues/3

Sending category, description and session id to splunk and generating the start batch file


src/PerformanceTests/Tests/Categories/Base.cs
```diff
@@ -95,11 +101,15 @@ static void DeleteOtherPlatformHost(Permutation permutation)
...
+    }
+    
+    [SetUpFixture]
+    public class TestSessionInitializer
+    {
+        [SetUp]
+        public void Setup()
         {
-            SessionId = Guid.NewGuid().ToString();
+            Base.SessionId = DateTime.UtcNow.Ticks.ToString();
```
ramonsmits: Why share the same session ID across *all* test? How can I filter on a single run?

ramonsmits: Also, trying to understand this:

>Attribute used to mark a class that contains one-time SetUp and/or TearDown methods that apply to all the tests in a namespace or an assembly.

Does this mean a new session ID per fixture, or for *all* fixtures?

src/PerformanceTests/Tests/Statistics.cs
```diff
@@ -0,0 +1,130 @@
+﻿using System;
```
ramonsmits: Why is this moved to the test project? It already is becoming the root of all evil. Stuff should be moved out of it.

ramonsmits: Also, you didn't move / rename the file as it is mared as "Added" in git, loosing history on its change

src/PerformanceTests/NServiceBus6/NServiceBus6.csproj
```diff
@@ -80,10 +80,6 @@
     <Reference Include="System.Xml" />
   </ItemGroup>
   <ItemGroup>
-    <ProjectReference Include="..\Common\Utils.csproj">
```
ramonsmits: You Removed the Utils project from the solution, but kept the folder.

src/PerformanceTests/Tests/Categories/Base.cs
```diff
@@ -39,10 +39,14 @@ public virtual void GatedPublishRunner(Permutation permutation)
 
         void Tasks(Permutation permutation, [CallerMemberName] string memberName = "")
         {
-            permutation.Category = GetType().Name;
+            var fixture = GetType().GetCustomAttribute<TestFixtureAttribute>();
+            permutation.Category = fixture.Category;
+            permutation.Description = fixture.Description;
```
ramonsmits: What is the benefit of passing a description all the way to the NLog global context and add it to each and every line?

src/PerformanceTests/Tests/Categories/Base.cs
```diff
@@ -61,17 +65,19 @@ static void LaunchAndWait(Permutation permutation)
             var sessionIdArgument = string.Format(" --sessionId={0}", SessionId);
 
             var exe = new FileInfo(permutation.Exe);
+            var arguments = PermutationParser.ToArgs(permutation) + processIdArgument + sessionIdArgument;
```
dvdstelt: Can't you do something like
```
var permutationArguments = PermutationParser.ToArgs(permutation);
var arguments = string.Format("{0} {1} {2}", permutationArguments, processIdArgument, sessionIdArgument);
```
I'd had to look at it carefully to see what the intention was there. Because PermutationParser.ToArgs() feels like a list is being returned, not a single string. So that concatenation wasn't clear to me.

src/PerformanceTests/Tests/Permutations/PermutationParser.cs
```diff
@@ -10,17 +10,24 @@ public static class PermutationParser
 
         public static string ToArgs(Permutation instance)
         {
-            return "--tests:" + string.Join(Seperator, instance.Tests) + " --variables:" + ToString(instance);
+            return "--tests:" + string.Join(Seperator, instance.Tests) + 
+                   " --category:" + instance.Category +
+                   " --description:\"" + instance.Description + "\"" +
+                   " --variables:" + ToString(instance);
```
dvdstelt: Why a : here, but a = in the other arguments? The parser will handle both, but = seems more consistent?

src/PerformanceTests/Tests/Categories/Base.cs
```diff
@@ -61,17 +65,19 @@ static void LaunchAndWait(Permutation permutation)
...
 
-            var pi = new ProcessStartInfo(exe.FullName, PermutationParser.ToArgs(permutation) + processIdArgument + sessionIdArgument)
+            var pi = new ProcessStartInfo(exe.FullName)
             {
+                Arguments = arguments,
```
ramonsmits: Don't use the arguments property while there is constructor overload that accepts arguments.

src/PerformanceTests/Tests/Permutations/PermutationParser.cs
ramonsmits: @dvdstelt The = is used for the variable value, not for the values of the arguments.

```
/tests:MySuperTest /variables:Audit=Off;DTC=On
```

ramonsmits:
So instead of moving stuff to Tests I would be in favor of moving stuff to Utils. The Utils should probably be renamed to something like Core. The current dependency chain where Tests is the root was a design flaw.

src/PerformanceTests/Tests/Categories/Base.cs
HEskandari: This is to generate a build ID if you will, where we run all the tests. To have a Test or Fixture ID we can have the same approach but generate them per test. What I wanted to do here was to separate latest 'batch' of tests from all the others builds/test run on the same day or previous days.

HEskandari: Also you can filter them by category and description + latest run. e.g. show me the test result of the latest build where test category is performance or test description is 'Audit forwarding On / Off'.

src/PerformanceTests/Tests/Statistics.cs
HEskandari: Need to read stuff off Permutation object. So this can go where that is. There was hardly any benefit of having a project with just Statistics in it.

src/PerformanceTests/Tests/Categories/Base.cs
HEskandari: We could use the test name instead, but the point was to use it as a filter if we need to filter out the result.

PR #50, opened 2016-04-14
Azure persistence

ramonsmits:
Changes:

- Azure persistence configuration added.
- MSMQ configuration string stored in connection string config file.
- NotSupportedExceptions shown as Inconclusive in NUnit runner.

The rest are fixes.


src/PerformanceTests/Persistence.V5.Azure_v6/AzureProfile.cs
```diff
@@ -1,9 +1,22 @@
...
-public class AzureProfile : IProfile
+public class AzureProfile : IProfile, INeedPermutation
 {
-    public void Configure(BusConfiguration busConfiguration)
+    public Permutation Permutation { private get; set; }
+
+    public void Configure(BusConfiguration cfg)
     {
-        busConfiguration.UsePersistence<AzureStoragePersistence>();
+        if (Permutation.OutboxMode == Outbox.On) throw new NotSupportedException("Outbox is not supported with Azure storage.");
```
dvdstelt: If INeedPermutation is set and Permutation is null, throw exception?

`if (this is INeedPermutation && Permutation == null) throw new InvalidOperationException("Cowabunga dude, you forgot to set Permutation property.");`

dvdstelt: Oh wait, and then in BaseRunner obviously...

src/PerformanceTests/Tests/Categories/Base.cs
```diff
@@ -84,9 +84,12 @@ static void LaunchAndWait(Permutation permutation)
                 {
                     p.Kill();
                     Assert.Fail("Killed!");
```
dvdstelt: Can't this be a little more descriptive? As in, I waited x-seconds and the process still seems to be working. Make it complete faster, please...

src/PerformanceTests/Utils/AssemblyScanner.cs
```diff
@@ -43,7 +43,7 @@ public static IEnumerable<Assembly> GetAssemblies()
                 version = ex.Message;
             }
 
-            l.InfoFormat("Loaded: {0} ({1})", name, version);
+            l.Info("Loaded: {0} ({1})", name, version);
```
dvdstelt: I don't get this? It doesn't use the parameters, does it?

src/PerformanceTests/Utils/QueueUtils.cs
```diff
@@ -0,0 +1,30 @@
+﻿using System;
+
+using System.Messaging;
+
+static class MsmqUtils
```
dvdstelt: Can be removed, isn't used anymore

src/PerformanceTests/Utils/ReturnCodes.cs
```diff
@@ -0,0 +1,5 @@
+﻿public enum ReturnCodes
+{
+    OK = 0,
+    NotSupported = 1000
```
dvdstelt: Why 1000? Why not 1337? :stuck_out_tongue_closed_eyes:

src/PerformanceTests/Utils/AssemblyScanner.cs
ramonsmits: Moved from NServiceBus logging to NLog, Nlog has a different method signature.

src/PerformanceTests/Utils/QueueUtils.cs
ramonsmits: I'll keep it in because we need logic to cleanup / remove the queues after a run.

src/PerformanceTests/Utils/ReturnCodes.cs
ramonsmits: @dvdstelt Your request is my command ;-)

src/PerformanceTests/Persistence.V5.Azure_v6/AzureProfile.cs
ramonsmits: You would not check infrastructural plumbing.

PR #52, opened 2016-04-14
Distribution

ramonsmits:
**Please merge #50 first as this branch is forked from *azure-persistence***

Adds distributor and sender-side distribution.

PR #53, opened 2016-04-14
GCServer patching

ramonsmits:
This patches the app.config and set the GCServer enabled attribute to true if the permutation used the server GC.

**Please merge #52 first, as this branch is forked from that.**


ramonsmits:
Wrong branch... master :(

PR #54, opened 2016-04-15
gcServer patching

ramonsmits:
- Patches the gcServer configuration value

Also

- Chatty INFO loggers filtered
- Cleanup logic to have a fallback when you have a folder open in explorer or the cmd.


WilliamBZA:
:+1: looks good

PR #61, opened 2016-04-18
Async api

ramonsmits:
- Scenarios don't need to write their own V5/V6 SendLocal, Publish, its abstracted via the `ISession` interface
- Parallelism extracted to helper class and refactored with @danielmarbach his implementation.

### DIFF
--- a/src/PerformanceTests/Common/Scenarios/BaseRunner.cs
+++ b/src/PerformanceTests/Common/Scenarios/BaseRunner.cs
@@ -20,12 +20,6 @@ public abstract class BaseRunner : IConfigurationSource, IContext
 {
     readonly ILog Log = LogManager.GetLogger("BaseRunner");
 
-#if Version5
-    protected IBus EndpointInstance { get; private set; }
-#else
-    IEndpointInstance EndpointInstance { get; set; }
-#endif
-
     public Permutation Permutation { get; private set; }
     public string EndpointName { get; private set; }
     protected ISession Session { get; private set; }
@@ -44,8 +38,7 @@ public abstract class BaseRunner : IConfigurationSource, IContext
 
         CreateSeedData();
 
-        EndpointInstance = CreateEndpoint();
-        Session = new Session(EndpointInstance);
+        CreateEndpoint();
 
         try
         {
@@ -60,11 +53,7 @@ public abstract class BaseRunner : IConfigurationSource, IContext
         }
         finally
         {
-#if Version5
-            using (EndpointInstance) { }
-#else
-            EndpointInstance.Stop().GetAwaiter().GetResult();
-#endif
+            Session.Close();
         }
     }
 
@@ -84,7 +73,7 @@ public abstract class BaseRunner : IConfigurationSource, IContext
     {
     }
 
-    private void CreateSeedData()
+    void CreateSeedData()
     {
         var seedCreator = this as ICreateSeedData;
         if (seedCreator == null) return;
@@ -93,7 +82,7 @@ public abstract class BaseRunner : IConfigurationSource, IContext
 
         CreateQueues();
 
-        var sendonlyInstance = CreateSendOnlyEndpoint();
+        CreateSendOnlyEndpoint();
 
         try
         {
@@ -102,17 +91,13 @@ public abstract class BaseRunner : IConfigurationSource, IContext
                 if (i % 5000 == 0)
                     Log.InfoFormat("Seeded {0} messages.", i);
 
-                ((ICreateSeedData)this).SendMessage(sendonlyInstance);
+                ((ICreateSeedData)this).SendMessage(Session);
             });
             Log.InfoFormat("Seeded total of {0} messages.", seedCreator.SeedSize);
         }
         finally
         {
-#if Version5
-            using (sendonlyInstance) { }
-#else
-            sendonlyInstance.Stop().GetAwaiter().GetResult();
-#endif
+            Session.Close();
         }
     }
 
@@ -121,17 +106,17 @@ public abstract class BaseRunner : IConfigurationSource, IContext
     {
         var configuration = CreateConfiguration();
         configuration.PurgeOnStartup(true);
-        var createQueuesBus = Bus.Create(configuration).Start();
-        createQueuesBus.Dispose();
+        using (Bus.Create(configuration).Start()){}
     }
 
-    ISendOnlyBus CreateSendOnlyEndpoint()
+    void CreateSendOnlyEndpoint()
     {
         var configuration = CreateConfiguration();
-        return Bus.CreateSendOnly(configuration);
+        var instance = Bus.CreateSendOnly(configuration);
+        Session = new Session(instance);
     }
 
-    IBus CreateEndpoint()
+    void CreateEndpoint()
     {
         var configuration = CreateConfiguration();
         configuration.EnableFeature<NServiceBus.Performance.SimpleStatisticsFeature>();
@@ -142,13 +127,12 @@ public abstract class BaseRunner : IConfigurationSource, IContext
         else
             configuration.PurgeOnStartup(true);
 
-        return Bus.Create(configuration).Start();
+        Session = new Session(Bus.Create(configuration).Start());
     }
 
-    private bool QueuesWerePurgedWhenSeedingData()
+    bool QueuesWerePurgedWhenSeedingData()
     {
-        if (this is ICreateSeedData) return true;
-        return false;
+        return this is ICreateSeedData;
     }
 
     BusConfiguration CreateConfiguration()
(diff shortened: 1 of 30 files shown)


src/PerformanceTests/Utils/TaskHelper.cs
```diff
@@ -0,0 +1,25 @@
...
+using System.Collections.Generic;
+using System.Threading.Tasks;
+
+public class TaskHelper
+{
+    static readonly int cores = Environment.ProcessorCount;
+
+    public static async Task ParallelFor(int count, Func<Task> action)
+    {
+        // If we have a batch of 10,000 and have 10 cores we do 10 iterations of 1,000 message chunks.
```
ramonsmits: @danielmarbach Was it really your intention to have this behavior? Super large sets would basically have the same behavior.

src/PerformanceTests/NServiceBus6/Session.cs
```diff
@@ -0,0 +1,27 @@
...
+    readonly IEndpointInstance instance;
+
+    public Session(IEndpointInstance instance)
+    {
+        this.instance = instance;
+    }
+
+    public Task Send(object message)
+    {
+        return instance.Send(message);
```
ramonsmits: @danielmarbach This is allowed correct? It is not required to do:

```c#
public async Task Send(object message)
{
    await instance.Send(message);
}
```

It feels strange to await when you can just return but I don't know if this is bad practice and if it is, why its so.

src/PerformanceTests/Utils/TaskHelper.cs
danielmarbach: Exactly what I said on the call. It was just a first hack. Not
sophisticated as you pointed out. The problem is that depending on the
latency and number of tasks it is better to chunk and sometimes not. I'm
fine if you guys say: let's not chunk

src/PerformanceTests/NServiceBus6/Session.cs
danielmarbach: I suggest you watch my webinar series ;)

In this case it is better to return. No statemachine creation = less memory
and no context capturing

ramonsmits: @danielmarbach Thanks!

src/PerformanceTests/Utils/TaskHelper.cs
ramonsmits: @danielmarbach I'm going to test it with all transports, see what works best.

src/PerformanceTests/Common/Scenarios/BaseRunner.cs
```diff
@@ -19,11 +19,12 @@ public abstract class BaseRunner : IConfigurationSource, IContext
 #if Version5
     protected IBus EndpointInstance { get; private set; }
 #else
-    protected IEndpointInstance EndpointInstance { get; private set; }
+    IEndpointInstance EndpointInstance { get; set; }
```
danielmarbach: Why is v5 protected in v6 not?

src/PerformanceTests/Common/Scenarios/GatedPublish/GatedPublishRunner.cs
```diff
@@ -34,10 +34,7 @@ partial class GatedPublishRunner : LoopRunner, IConfigureUnicastBus
                     X.Reset(batchSize);
 
                     var d = Stopwatch.StartNew();
-
-                    var sends = new Task[X.InitialCount];
-                    for (var i = 0; i < X.InitialCount; i++) sends[i] = Publish(new Event());
-                    await Task.WhenAll(sends);
+                    await TaskHelper.ParallelFor(X.InitialCount, () => Session.Publish(new Event()));
```
danielmarbach: Would you want to apply `ConfigureAwait(false)` consistently?

src/PerformanceTests/Common/Scenarios/GatedSendLocal/GatedSendLocalRunner.cs
```diff
@@ -33,9 +33,7 @@ partial class GatedSendLocalRunner : LoopRunner
 
                 var d = Stopwatch.StartNew();
 
-                var sends = new Task[X.InitialCount];
-                for (var i = 0; i < X.InitialCount; i++) sends[i] = SendLocal(CommandGenerator.Create());
-                await Task.WhenAll(sends);
+                await TaskHelper.ParallelFor(X.InitialCount, () => Session.SendLocal(CommandGenerator.Create()));
```
danielmarbach: Would you want to apply `ConfigureAwait(false)` consistently?

src/PerformanceTests/Common/Scenarios/SendLocalOneOnOne/SendLocalOneOnOneRunner.cs
```diff
@@ -16,9 +16,7 @@ partial class SendLocalOneOnOneRunner : BaseRunner
 
     protected override void Start()
```
danielmarbach: Why is start not async?

src/PerformanceTests/NServiceBus5/Session.cs
```diff
@@ -0,0 +1,27 @@
...
+    readonly IBus instance;
+
+    public Session(IBus instance)
+    {
+        this.instance = instance;
+    }
+
+    public async Task Send(object message)
+    {
+        instance.Send(message);
```
danielmarbach: I would advice to remove `async` and return `Task.CompletedTask`

src/PerformanceTests/NServiceBus5/Session.cs
```diff
@@ -0,0 +1,27 @@
...
+    }
+
+    public async Task Publish(object message)
```
danielmarbach: I would advice to remove `async` and return `Task.CompletedTask`

src/PerformanceTests/NServiceBus5/Session.cs
```diff
@@ -0,0 +1,27 @@
...
+    {
+        instance.Publish(message);
+    }
+
+    public async Task SendLocal(object message)
```
danielmarbach: I would advice to remove `async` and return `Task.CompletedTask`

src/PerformanceTests/Common/Scenarios/GatedPublish/GatedPublishRunner.cs
dvdstelt: I understood that this is more for frameworks/libraries like NSB, not so much for your own code?

danielmarbach: We do it in the acceptance testing framework as well to remove all potential side effects

src/PerformanceTests/Common/Scenarios/BaseRunner.cs
```diff
@@ -100,53 +90,40 @@ private void CreateSeedData()
                 if (i % 5000 == 0)
                     Log.InfoFormat("Seeded {0} messages.", i);
 
-                ((ICreateSeedData)this).SendMessage(sendonlyInstance);
+                ((ICreateSeedData)this).SendMessage(Session);
```
danielmarbach: SendMessage returns a Task. You have to await it or use GetAwaiter().GetResult()

ramonsmits: Awesome find @danielmarbach !

ramonsmits: @danielmarbach Are you using some kind of extension that mentions that?

src/PerformanceTests/Common/Scenarios/SendLocalOneOnOne/SendLocalOneOnOneRunner.cs
ramonsmits: Because the code currently is pretty clean and adding async ceremony for micro-optimization is not needed.

danielmarbach: See and that's the root problem of the previous comments. When you have a
code base which moxed void and task returning APIs you will run into issues
I found above. And async call inside a void method is treated as
fire&forget. The best practice is async all the way and that for a good
reason. About your cleannes and microptimization: it is not about
microoptimization. My point is entirely about cleanness. Because currently
it is not clean in terms of async code flow

ramonsmits: You are right about the async/sync workflow. That is something to consider. @WilliamBZA @dvdstelt Your opinion?

dvdstelt: Let's go all out, as @danielmarbach says. Worst case scenario is that we might learn something ;-)

PR #63, opened 2016-04-19
Enabling encryption

hmemcpy:


src/PerformanceTests/Common/Scenarios/Receive/ReceiveRunner.V5.cs
```diff
@@ -8,19 +9,31 @@ partial class ReceiveRunner : ICreateSeedData
 {
     public int SeedSize { get; set; } = 50000;
     
-    public partial class Handler : IHandleMessages<Command>
+    public partial class Handler : IHandleMessages<Command>, IHandleMessages<EncryptedCommand>
```
ramonsmits: Don't re-use an existing scenerio, create a new one.

src/PerformanceTests/NServiceBus5/App.config
```diff
@@ -67,10 +67,6 @@
         <assemblyIdentity name="Newtonsoft.Json" publicKeyToken="30ad4fe6b2a6aeed" culture="neutral" />
         <bindingRedirect oldVersion="0.0.0.0-4.5.0.0" newVersion="4.5.0.0" />
       </dependentAssembly>
-      <dependentAssembly>
```
ramonsmits: Why is the NLog assembly redirect removed?

src/PerformanceTests/Common/Scenarios/Receive/ReceiveRunner.V5.cs
dvdstelt: +1, the baserunner and looprunner are there to help create different scenarios. It's just a few lines of code to create a new one.

PR #72, opened 2016-04-25
Improve intent of session

dvdstelt:
The current implementation hides away the intent of `ISession` inside the runners. Developer has to figure out how abstract method SendMessage() has to send or publish messages. 

With `ISession` as parameter into the SendMessage method, the intent becomes clear that this object has to be used by developer, creating a new scenario.

### DIFF
--- a/src/PerformanceTests/Common/Scenarios/LoopRunner.cs
+++ b/src/PerformanceTests/Common/Scenarios/LoopRunner.cs
@@ -18,12 +18,12 @@ abstract class LoopRunner : BaseRunner
     CancellationTokenSource stopLoop { get; set; }
     bool Shutdown { get; set; }
     int BatchSize { get; set; } = 16;
-    protected abstract Task SendMessage();
+    protected abstract Task SendMessage(ISession session);
 
-    protected override void Start()
+    protected override void Start(ISession session)
     {
         stopLoop = new CancellationTokenSource();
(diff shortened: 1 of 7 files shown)


src/PerformanceTests/Common/Scenarios/LoopRunner.cs
```diff
@@ -57,7 +59,7 @@ protected override void Stop()
                     countdownEvent.Reset(BatchSize);
                     var batchDuration = Stopwatch.StartNew();
 
-                    await TaskHelper.ParallelFor(BatchSize, SendMessage);
+                    await TaskHelper.ParallelFor(BatchSize, () => SendMessage(Session));
```
WilliamBZA: Can we change the signature of the `Loop` method to also accept the `ISession` as a parameter? That way we can completely drop the class level variable completely.

dvdstelt: @WilliamBZA Fixed it

src/PerformanceTests/Common/Scenarios/SendLoop/BaseLoop.cs
```diff
@@ -57,7 +60,7 @@ protected override void Stop()
 
                     var batchDuration = Stopwatch.StartNew();
 
-                    await Batch(BatchSize, SendMessage);
+                    await Batch(BatchSize, () => SendMessage(Session));
```
WilliamBZA: Also this one :trollface:

PR #88, opened 2016-04-29
[WIP] Add FLR to Azure Storage Queues to bypass throttling 500 errors

WilliamBZA:
We are getting intermittent 500 errors from Azure. This appears to be Azure throttling us.

https://azure.microsoft.com/en-us/documentation/articles/storage-scalability-targets/

By enabling FLR and SLR the retries _should_ allow the messages to process correctly.

This does however allow the chance that the tests will now fail due to running over the test timecap.

### DIFF
--- a/src/PerformanceTests/Transport.V5.AzureStorageQueues.V6/AzureStorageQueueProfile.cs
+++ b/src/PerformanceTests/Transport.V5.AzureStorageQueues.V6/AzureStorageQueueProfile.cs
@@ -37,6 +37,19 @@ class FlrConfig : IConfigurationSource
             return flrConfig as T;
         }
 
+        // To provide SLR Config
+        if (typeof(T) == typeof(SecondLevelRetriesConfig))
+        {
+            SecondLevelRetriesConfig slrConfig = new SecondLevelRetriesConfig
+            {
+                Enabled = true,
+                NumberOfRetries = 2,
+                TimeIncrease = TimeSpan.FromSeconds(10)
+            };
+
+            return slrConfig as T;
+        }
+
         // To in app.config for other sections not defined in this method, otherwise return null.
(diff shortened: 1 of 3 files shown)


dvdstelt:
It's good that it's configured in the ASQ profile, but how does this effect message throughput? The exception self will slow processing down, but the FLR will probably increase it, as it doesn't need to get the message from the transport 2 times in a row.

And you also configured SLR? What does that mean for the transport and performance?

WilliamBZA:
@dvdstelt Yes, you are 100% correct. Good call, this is likely not the correct solution. But let's use this PR as a place of discussion.

We _are_ going to get 500 errors as we saturate the Azure requests (occasionally). If our tests are doing anything reasonably performant it must happen from time to time. A symptom of this is that our tests will fail intermittently because if the unhandled application exception.

What does this mean though? Well I think that depends on how we decide to treat these 500's.

If we fix the error, and follow the MS recommended approach of retrying the request after a small delay (FLR + SLR), we're skewing the results by artificially reducing the number of messages processed in a given timeframe.

We can ignore the error and let 500's still pass, but stop executing at that point in time. Which again means we're skewing the results. We could also be getting the 500's during seeding in which case we won't actually even get a result for that test case.

Or we could put a `try...catch` around the entire test and simply keep retrying that test permutation case until we don't get a 500 (possibly with a maximum retry count of 3?).

Any other options I haven't through of?

/cc @ramon @igal

WilliamBZA:
Leaving this for now.

dvdstelt:
@WilliamBZA Was it for receiving messages, sending messages or both?

WilliamBZA:
@dvdstelt I'm not sure anymore. Both contribute to the throttling limit, so it would probably be a combination of both.

PR #107, opened 2016-05-11
WIP: Additional message sizes

ramonsmits:
Additional message sizes to make comparison easier with results from @danielmarbach

### DIFF
--- a/src/PerformanceTests/Utils/Variables/MessageSize.cs
+++ b/src/PerformanceTests/Utils/Variables/MessageSize.cs
@@ -4,7 +4,14 @@ namespace Variables
     {
         Tiny = 0,
         Small = 1024,
-        Medium = 10 * 1024,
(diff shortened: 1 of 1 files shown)


ramonsmits:
@Particular/endtoend-maintainers I'm not very happy with the enum value names. Any suggestions for improvements?

ramonsmits:
The largest additional message payload that we are currently testing is 100KB. @danielmarbach  is also testing with several larger messages.

To my knowledge there is no real limit for SQL Server and RabbitMQ, a 4MB limit for MSMQ and a 1MB limit for ASB premium, 256KB for ASB, 64KB for ASQ.

Is it worth testing performance differences between V5 and V6 for such large messages if the transport supports it?

ramonsmits:
Including this in the current form will make the test run much longer than it already is without providing much valuable feedback.

PR #108, opened 2016-05-11
WIP: Seeding now using same maximum concurrency as receiving.

ramonsmits:
Seeding was not restricted in its maximum concurrency. This restricts concurrency to the seeding is more aligned in throughput performance with the receiving.

### DIFF
--- a/src/PerformanceTests/Common/Scenarios/BaseRunner.cs
+++ b/src/PerformanceTests/Common/Scenarios/BaseRunner.cs
@@ -88,7 +88,12 @@ public abstract class BaseRunner : IConfigurationSource, IContext
 
             var count = 0L;
 
-            Parallel.ForEach(IterateUntilFalse(() => !cts.Token.IsCancellationRequested), b =>
+            var po = new ParallelOptions
+            {
+                MaxDegreeOfParallelism = ConcurrencyLevelConverter.Convert(Permutation.ConcurrencyLevel)
+            };
+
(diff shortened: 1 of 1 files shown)


ramonsmits:
@Particular/endtoend-maintainers The gated scenarios currently uses a task-based approach for the sending outside of a message handler context.

Following options:

- Don't use task based approach but also a Parallel.For
- Use a semaphore to limit concurrency during sending
- Ignore the fact that sending is unbounded in concurrency


What do you think?

src/PerformanceTests/Common/Scenarios/BaseRunner.cs
```diff
@@ -88,7 +88,12 @@ void CreateSeedData()
 
             var count = 0L;
 
-            Parallel.ForEach(IterateUntilFalse(() => !cts.Token.IsCancellationRequested), b =>
+            var po = new ParallelOptions
+            {
+                MaxDegreeOfParallelism = ConcurrencyLevelConverter.Convert(Permutation.ConcurrencyLevel)
```
johnsimons: I am pretty sure this setting is actually by default the number of cores, so if `Permutation.ConcurrencyLevel` is higher then the number of cores, you will actually see a degradation in performance, see https://msdn.microsoft.com/en-us/library/system.threading.tasks.paralleloptions.maxdegreeofparallelism(v=vs.110).aspx#Anchor_2

ramonsmits:
Made obsolete by #111

PR #117, opened 2016-05-20
Testing saga get by correlation property

tmasternak:
Testing saga retrieval by `CorrelationProperty`.


tmasternak:
ping @Particular/endtoend-maintainers

SimonCropp:
shouldnt we instead disable lazy loading for the tests?

tmasternak:
@SimonCropp good point. The reasons that I've put that in the first place where:
 * we could disable lazy-loading on the entity level but  `Persister` does not have any API to expose that
 * we could use the `ISatelessSession` instead of `ISession` but in v7 of `Persister` needs the latter  to do initialization

That being said I did a bit more research and found out that there is `NHibernateUtil.Initialize` method which does initialization for proxy object. Unfortunately it does not initialize child collections :/ so we can't use it in our scenario.

tmasternak:
Closing this one. It has been merged into https://github.com/Particular/EndToEnd/pull/119

PR #118, opened 2016-05-23
Optimizing app domian creation

tmasternak:
This change enables reuse of `AppDomain`s between test cases making test run order of magnitude shorter.

ping @Particular/endtoend-maintainers


src/NHPersistenceCompatibilityTests/PersistenceCompatibilityTests/PersisterProvider.cs
```diff
@@ -0,0 +1,62 @@
...
+
+        public void Dispose()
+        {
+            foreach (var appDomainDescriptor in appDomainDescriptors)
+            {
+                appDomainDescriptor.Dispose();
+
+                new FileInfo(appDomainDescriptor.ProjectAssemblyPath)
+                    .Directory
+                    ?.Delete(true);
```
HEskandari: I've had this failing occasionally, specially when running with Resharper test runner. Should we wrap the delete in a try-catch?

HEskandari:
@tmasternak Looks good :+1:

src/NHPersistenceCompatibilityTests/PersistenceCompatibilityTests/PersisterProvider.cs
tmasternak: I think that it happens only in Debug under VS. The reason I think is that VS debugger holds handle to `NServiceBus.NHibernate.Tests.*` files even after the app domain as unloaded. 

I do not see this behavior in `Release` mode so we should be good when running this under TC. I would keep it the way it is and if we see problems in `Relese` we can fix it.

HEskandari: I *think* it happened to me today when running the test, not just debugging them. I'll have an eye on that tomorrow, see if it happens again.

PR #120, opened 2016-05-24
Cleaning database before test runs

HEskandari:
Cleans the database tables when test starts. This would allow us access the tables (for diagnostics) after a test fails.

### DIFF
--- a/src/NHPersistenceCompatibilityTests/Common/NHibernateConnectionInfo.cs
+++ b/src/NHPersistenceCompatibilityTests/Common/NHibernateConnectionInfo.cs
@@ -4,8 +4,8 @@ namespace Common
 {
     public class NHibernateConnectionInfo
     {
-        const string Dialect = "NHibernate.Dialect.MsSql2012Dialect";
-        const string ConnectionString = @"Data Source=localhost\SQLEXPRESS;Initial Catalog=persistencetests;Integrated Security=True";
+        public const string Dialect = "NHibernate.Dialect.MsSql2012Dialect";
+        public const string ConnectionString = @"Data Source=localhost\SQLEXPRESS;Initial Catalog=persistencetests;Integrated Security=True";
 
         public static IDictionary<string, string> Settings = new Dictionary<string, string>
         {
--- a/src/NHPersistenceCompatibilityTests/PersistenceCompatibilityTests/Database.cs
+++ b/src/NHPersistenceCompatibilityTests/PersistenceCompatibilityTests/Database.cs
@@ -1,0 +1,32 @@
+﻿namespace PersistenceCompatibilityTests
+{
+    using System.Data.SqlClient;
+    using Common;
+
+    public class Database
+    {
+        public static void Cleanup()
+        {
+            var connectionString = NHibernateConnectionInfo.ConnectionString;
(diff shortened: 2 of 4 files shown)


src/NHPersistenceCompatibilityTests/Common/NHibernateConnectionInfo.cs
```diff
@@ -4,8 +4,8 @@ namespace Common
 {
     public class NHibernateConnectionInfo
     {
-        const string Dialect = "NHibernate.Dialect.MsSql2012Dialect";
-        const string ConnectionString = @"Data Source=localhost\SQLEXPRESS;Initial Catalog=persistencetests;Integrated Security=True";
+        public const string Dialect = "NHibernate.Dialect.MsSql2012Dialect";
+        public const string ConnectionString = @"Data Source=localhost\SQLEXPRESS;Initial Catalog=persistencetests;Integrated Security=True";
```
ramonsmits: @HEskandari Instead of making this public, wouldn't it be better to put this in the configuration connection strings?

src/NHPersistenceCompatibilityTests/PersistenceCompatibilityTests/Database.cs
```diff
@@ -0,0 +1,32 @@
...
+            var builder = new SqlConnectionStringBuilder(connectionString);
+            var initialCatalog = builder.InitialCatalog;
+            builder.InitialCatalog = "master";
+
+            using (var connection = new SqlConnection(builder.ConnectionString))
+            using (var command = connection.CreateCommand())
+            {
+                connection.Open();
+
+                var query = @"IF EXISTS(SELECT * from sys.databases where name = '{0}')
```
ramonsmits: This looks like if its TSQL specific. The Dialect property was made public. Does this only run for SQL Server? Shouldn't there be a dialect check?

HEskandari: Yeah, the scope of this issue is to run it on Sql only.

tmasternak: @ramonsmits yes this is supposed to run only on top of SQL Server

src/NHPersistenceCompatibilityTests/Common/NHibernateConnectionInfo.cs
tmasternak: What do you mean by `configuration connection strings`?

HEskandari: That'd be a bit tricky. Putting this into app.configs we have to put them into app.config of each project/version, which means more duplicates. Furthermore, this is executed using AppDomains which means more trouble loading the right app.config for a given AppDomain. The simplest approach would be just having one shared connection-string in code. Should we need to configure it, I'd go with environment variables instead of app.config approach.

HEskandari: @tmasternak I think Ramon means putting it in `ConnectionStrings` section of the config file.

PR #122, opened 2016-05-24
Make nuget the source for package versions

WilliamBZA:
Connects to https://github.com/Particular/PlatformDevelopment/issues/737

Most test permutations are passing except for almost all `v6.*.*` tests. There is a weird reflection error coming up.

Ping @tmasternak @HEskandari @weralabaj


tmasternak:
@WilliamBZA I'll add `WIP` prefix until we make it run.

tmasternak:
It looks that in version `6` the `IStorageSessionProvider` has changed. E.g. [earlier-version] (https://github.com/Particular/NServiceBus.NHibernate/blob/support-6.0/src/NServiceBus.NHibernate/SharedSession/IStorageSessionProvider.cs) and [later-version](https://github.com/Particular/NServiceBus.NHibernate/blob/support-6.1/src/NServiceBus.NHibernate/SharedSession/IStorageSessionProvider.cs). 

I'll try to come-up with solution and we can discuss.

tmasternak:
So it turned out that there where actually 3 versions of `IStorageSessionProvider`. I've added two projects to handle that. The tests seem to work now but some polishing is definitely needed.

/cc: @HEskandari

src/NHPersistenceCompatibilityTests/6.1.0/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,36 @@
+﻿using System.Reflection;
```
HEskandari: Don't think we need it. It was removed from the other projects (the whole AssemblyInfo file)

WilliamBZA:
Ready for review and merge now @tmasternak @HEskandari @weralabaj

src/NHPersistenceCompatibilityTests/Common/NugetHelper.cs
```diff
@@ -27,10 +30,25 @@ public IEnumerable<string> GetPossibleVersionsFor(string packageName, string min
 
         internal void DownloadPackageTo(string packageName, string version, string location)
         {
-            var repo = PackageRepositoryFactory.Default.CreateRepository(packageSource);
+            try
+            {
+                InstallPackageFromSource(packageSource, packageName, version, location);
+            }
+            catch (Exception e)
```
HEskandari: This catch-all and install from alternative source seems a bit random? any specific exception we can catch? In which condition might this fail?

src/NHPersistenceCompatibilityTests/PersistenceCompatibilityTests/UpdatingSagaTests.cs
```diff
@@ -125,17 +125,19 @@ public void can_fetch_composite_saga_updated_by_another_version(string sourceVer
...
+            NHibernatePackageVersions = nuget.GetPossibleVersionsFor("NServiceBus.NHibernate", "4.5.0");
+
             var cases = from va in NHibernatePackageVersions
                         from vb in NHibernatePackageVersions
                         select new object[] { va, vb };
 
             return cases.ToArray();
         }
 
+        private static IEnumerable<string> NHibernatePackageVersions;
```
HEskandari: Different Resharper settings? think we omit the default access modifiers.

HEskandari:
Not part of this PR, but the `Fetcher` stuff are generic enough to be moved out and shared. Any reason we repeat them in each version, e.g. 6.0 vs 6.10?

src/NHPersistenceCompatibilityTests/Common/NugetHelper.cs
tmasternak: @HEskandari this is caused by the fact that for beta-0003 `NServiceBus.NHiberante` is on nuget but `NServiceBus` dependency is only on myget. Unfortunately the exceptions throw here are very generic :/

src/NHPersistenceCompatibilityTests/PersistenceCompatibilityTests/UpdatingSagaTests.cs
WilliamBZA: Perhaps I should enable resharper before we look at resharper settings :stuck_out_tongue:

weralabaj:
> Not part of this PR, but the Fetcher stuff are generic enough to be moved out and shared. Any reason we repeat them in each version, e.g. 6.0 vs 6.10?

it's included in SharedDefinitions project, so in fact it's not repeated. Does that address your comment or you'd like to move it to a separate library project?

weralabaj:
I addressed Resharper's complaints.

@ramonsmits can you please review and merge?

src/NHPersistenceCompatibilityTests/Common/NugetHelper.cs
```diff
@@ -0,0 +1,54 @@
...
+using System;
+
+namespace Common
+{
+    public class NugetHelper
+    {
+        string packageSource;
+        string fallbackPackageSource;
+
+        public NugetHelper(string packageSource = "https://packages.nuget.org/api/v2", string fallbackPackageSource = "https://www.myget.org/F/particular/")
```
ramonsmits: What is the reason these values are in code? Also, next time we add a third argument?

ramonsmits:
I don't see something obviously wrong but I have no idea what is going on and why. I cannot really judge the changes.

As I see @WilliamBZA @tmasternak @HEskandari @weralabaj to have contributed and added this solution. If you are aligned then I don't have any issue with merging this. I don't consider myself to decide if this is correct from a behavioral or design perspective.

weralabaj:
Ok @WilliamBZA said he'll reply to your comment as soon as he's back home. Then he can ping you to merge.

src/NHPersistenceCompatibilityTests/Common/NugetHelper.cs
WilliamBZA: There shouldn't be a third argument. The thinking is that we should test the packages that are available to our users, so that means Nuget or MyGet currently. Ideally we would have preferred for Nuget to be the only source (it's a more properly curated release list than MyGet) but due to the nature of the release churn with beta's we need to have a backup.

Summary: I don't foresee a third argument adding any value in the future. In fact: I think even having 2 arguments is a code smell that we're using to work around current shortcomings.

I do however agree about the values in code. I shall move them to config, re-push and ping again.

WilliamBZA:
I've changed the code to load all nuget settings from the `C:\` drive.

That is currently the only hard-coded string, but I think that's fine. It will look for all nuget package sources defined in the machine and try each of those (starting with the official ones first).

@ramonsmits @tmasternak @HEskandari @weralabaj Take a look and let me know if there's anything you're uncomfortable with.

HEskandari:
I'm happy with the changes @WilliamBZA

weralabaj:
Couldn't that be loaded from app.config? Not a big deal since it'll be running on TC, where we can create whatever files we need, but app.config seems like a more natural location for me.

src/NHPersistenceCompatibilityTests/Common/NugetHelper.cs
```diff
@@ -2,25 +2,31 @@
...
-            // Todo: Load this package source from the machine config
-            // Todo: Support multiple package sources (e.g. Build server with Myget and Nuget)
-            this.packageSource = packageSource;
-            this.fallbackPackageSource = fallbackPackageSource;
+            LoadNugetSourcesFromConfig();
+        }
+
+        private void LoadNugetSourcesFromConfig()
+        {
+            var machineSettings = Settings.LoadDefaultSettings(new PhysicalFileSystem("C:\\"), null, null);
```
ramonsmits: Not sure what happens here

    var machineSettings = Settings.LoadDefaultSettings(new PhysicalFileSystem("C:\\"), null, null);

When reading machine settings I alway think about `machine.config`

WilliamBZA:
The nuget settings use the default Nuget config hierarchy defined [here](https://docs.nuget.org/consume/nuget-config-file), it just starts it's search at `C:\`. This means it will look in the `%APPDATA%\NuGet\NuGet.Config` config file as well as the local project's config override. This follows the standard hierarchy that the Nuget Command Line does - which will make debugging package errors a whole lot easier than if the values were stored in app.config.

src/NHPersistenceCompatibilityTests/Common/NugetHelper.cs
WilliamBZA: It's the machine-wide Nuget config settings. The same as if you ran Nuget from the CLI. A better name might be `cliSettings`?

weralabaj:
Fair enough, to be honest I think it was better hard-coded, but we can leave that out for the final cleanup and just get that one merged.

ping @WilliamBZA @ramonsmits

WilliamBZA:
I'm happy either way :)

src/NHPersistenceCompatibilityTests/Common/NugetHelper.cs
ramonsmits: @WilliamBZA I see, those types are from the Nuget assembly.

Would any of the paths apply mentioned here: https://docs.nuget.org/consume/nuget-config-file

So that would be `%APPDATA%\NuGet\NuGet.Config` or is it a different kind of path?

If that is the path then please adjust it. If this is a different kind of path then leave it.

tmasternak:
WOOT :)

Issue #132, opened 2016-05-31
Update NUnit for performance tests solution


ramonsmits:

ramonsmits:
NUnit 3.x is now working locally but it does not work on the build server:

```
NUnit error running tests in 'C:\BuildAgent\work\fa5f215bc71fc336\perftests\Tests.dll' assembly
Loading assembly is compiled for v4.0.30319, MSIL
NUnit runner runtime is v4.0.30319, x86
NUnit.Core.UnsupportedFrameworkException: Skipped loading assembly Tests because it references an unsupported version of the nunit.framework, 3.2.1.0

Server stack trace: 
   at NUnit.Core.Builders.TestAssemblyBuilder.GetCandidateFixtureTypes(Assembly assembly, String ns)
   at NUnit.Core.Builders.TestAssemblyBuilder.GetFixtures(Assembly assembly, String ns)
   at NUnit.Core.Builders.TestAssemblyBuilder.Build(String assemblyName, Boolean autoSuites)
   at NUnit.Core.Builders.TestAssemblyBuilder.Build(String assemblyName, String testName, Boolean autoSuites)
   at NUnit.Core.TestSuiteBuilder.Build(TestPackage package)
   at NUnit.Core.SimpleTestRunner.Load(TestPackage package)
   at NUnit.Core.ProxyTestRunner.Load(TestPackage package)
   at NUnit.Core.ProxyTestRunner.Load(TestPackage package)
   at NUnit.Core.RemoteTestRunner.Load(TestPackage package)
   at System.Runtime.Remoting.Messaging.StackBuilderSink._PrivateProcessMessage(IntPtr md, Object[] args, Object server, Object[]& outArgs)
   at System.Runtime.Remoting.Messaging.StackBuilderSink.SyncProcessMessage(IMessage msg)

Exception rethrown at [0]: 
   at System.Runtime.Remoting.Proxies.RealProxy.HandleReturnMessage(IMessage reqMsg, IMessage retMsg)
   at System.Runtime.Remoting.Proxies.RealProxy.PrivateInvoke(MessageData& msgData, Int32 type)
   at NUnit.Core.TestRunner.Load(TestPackage package)
   at JetBrains.TeamCity.NUnitLauncher.NUnitRunner.LoadTest(String testAssembly, String testName, TestRunner runner) in c:\BuildAgent\work\ad31cec0a1b0f083\src\NUnitLauncher\src\NUnit-2.6\NUnitRunner.cs:line 53
[...]

PR #134, opened 2016-06-01
Performance tests NUnit 3.x fixes

ramonsmits:
Solves #132

### DIFF
--- a/src/PerformanceTests/Tests/Categories/Base.cs
+++ b/src/PerformanceTests/Tests/Categories/Base.cs
@@ -133,14 +133,4 @@ namespace Categories
             }
         }
     }
-
-    [SetUpFixture]
-    public class TestSessionInitializer
-    {
-        [SetUp]
-        public void Setup()
-        {
-            Base.SessionId = DateTime.UtcNow.Ticks.ToString();
-        }
-    }
 }
--- a/src/PerformanceTests/Tests/Categories/TestSessionInitializer.cs
+++ b/src/PerformanceTests/Tests/Categories/TestSessionInitializer.cs
@@ -1,0 +1,15 @@
+namespace Categories
+{
+    using System;
+    using NUnit.Framework;
+
+    [SetUpFixture]
+    public class TestSessionInitializer
+    {
+        [OneTimeSetUp]
+        public void Setup()
+        {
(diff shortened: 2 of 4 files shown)


src/PerformanceTests/Tests/Categories/Base.cs
```diff
@@ -133,14 +133,4 @@ static void DeleteOtherPlatformHost(Permutation permutation)
             }
         }
     }
-
-    [SetUpFixture]
-    public class TestSessionInitializer
-    {
-        [SetUp]
```
WilliamBZA: Should this not be `[OneTimeSetup]` instead of replacing it with a static initialization?

I mean, the static init works fine, so I guess this is just a style question.

ramonsmits: If it was style, I would not make the change. The previous code resulted in :

> OneTimeSetUp: SetUpAttribute attribute not allowed in a SetUpFixture

This resolved the run-time error.

WilliamBZA: With V3 I can run this without any errors or warnings:

Note the change from `[Setup]` to `[OneTimeSetUp]`

```CSharp
[SetUpFixture]
public class SetupFixtu
{
    [OneTimeSetUp]
    public void setup()
    {
        
    }
}
```

But like I said, we're only doing a simple initialization of a single variable. Not a major issue either way :smile:

src/PerformanceTests/Tests/Categories/TestSessionInitializer.cs
```diff
@@ -0,0 +1,15 @@
...
+    using System;
+    using NUnit.Framework;
+
+    [SetUpFixture]
+    public class TestSessionInitializer
+    {
+        [OneTimeSetUp]
+        public void Setup()
+        {
+            Base.SessionId = DateTime.UtcNow.Ticks.ToString();
```
SimonCropp: Would this be simpler to init it in a static field?

ramonsmits: @SimonCropp LOL, I refactored it to that at first. Then changed it back as to keep current code similar.

PR #136, opened 2016-06-01
NHibernateProfile setup using wrong permutation value

ramonsmits:
Setup was using incorrect permutation value (Transport) to retrieve connection string, now set to persister.


WilliamBZA:
:+1: LGTM!

PR #142, opened 2016-06-06
Pubsub issues

ramonsmits:
Native pubsub got mis configured due to endpoint name differences between x86 / x64 and pubsub broken in V5 due to unsupported / broken scale out default when not using the azure host.

### DIFF
--- a/src/PerformanceTests/Common/Program.cs
+++ b/src/PerformanceTests/Common/Program.cs
@@ -13,10 +13,7 @@ namespace Host
     {
         static ILog Log;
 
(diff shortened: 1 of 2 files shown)


dvdstelt:
@ramonsmits What was the reason we initially used the appdomain name?

ramonsmits:
@dvdstelt I don't know, that was in the beginning when we didn't have a 64 and 32 bit host but then we also had V5 / V6 in the name. This causes issues with pubsub topology. For example, the same event send to 4 queues (v5/v6/x86/x64) but in the beginnen also to just 1.

dvdstelt:
So performance was increased?! :grin:

PR #143, opened 2016-06-06
Packages 20160603

ramonsmits:
Package update.

*Beta3 breaks pubsub, do not merge yet!*


SimonCropp:
@ramonsmits beta4 is out does that fix this?

ramonsmits:
@SimonCropp I will check just to be sure, thanks!

ramonsmits:
@SimonCropp Beta4 is in #145 but due to a breaking change (don't know exactly what..) I cannot compile the whole solution yet. However, Beta4 pubsub seems to be working again (/cc: @timbussmann )

PR #146, opened 2016-06-15
[WIP] SQL Transport ServiceControl Compatibility Tests

mikeminutillo:
Related To https://github.com/Particular/PlatformDevelopment/issues/833

Adds a test project that spins up a ServiceControl instance for each test case.

To get started you need the latest SC bits [from the build server](http://builds.particular.net/viewType.html?buildTypeId=ServiceControl_NewBuil). Download the zip asset, unblock it, and unzip it into `C:\Temp\ServiceControl`. This will eventually be imported as a build asset from the SC build. 

For each test case this will:
- Delete the SC Install folder
- Create the SC install folder
- Copy the raw SC files into the install folder
- Copy the SQL transport files into the install folder
- Write an `app.config` into the install folder (with the configured connection string)
- Spin up the SC instance 
- Wait for the API to start responding
  - This is where your test case runs
- Stop the SC instance


mikeminutillo:
@Particular/endtoend-maintainers we are at the end of the SQL Transport part of this but will be going deeper into the other transports one at a time. Should we get this one merged back to master and open a new one for any ongoing work or just keep working in this PR for now? We are isolated from the rest of the repo anyway with our own project/solution structure.

PR #147, opened 2016-06-17
Generate test cases based on implementations found

WilliamBZA:
Related To https://github.com/Particular/PlatformDevelopment/issues/833

Builds on top of https://github.com/Particular/EndToEnd/pull/146

* Changed the web requests to be async
* Generate the test cases based on the implementations if `ITransportDetails` in the assemblies
* Create an activator `Func<T>` to cater for different types of `ITransportDetails`

Depending on the differences between the versions of the transports, we may need to separate the implementations of the `ITransportDetails` into separate projects and have different ones for each version of the transports. (not in scope for now?)

ping @mikeminutillo @ramonsmits @gbiellem @HEskandari

### DIFF
--- a/src/PlatformCompatibilityTests/ServiceControlCompatibilityTests/Infrastructure/Sql/SqlScTest.cs
+++ b/src/PlatformCompatibilityTests/ServiceControlCompatibilityTests/Infrastructure/Sql/SqlScTest.cs
@@ -1,33 +1,41 @@
 ﻿namespace ServiceControlCompatibilityTests
 {
     using NUnit.Framework;
+    using System;
+    using System.Collections.Generic;
 
     abstract class SqlScTest
     {
-        ServiceControlInstance serviceControl;
-
-        [SetUp]
-        public void Setup()
+        Dictionary<Type, Func<ITransportDetails>> transportDetailActivations = new Dictionary<Type, Func<ITransportDetails>>
         {
-            var context = TestContext.CurrentContext;
-
-            // TODO: This is a build artifact from the latest SC build. It should be imported into the working folder of the build
-            var factory = new ServiceControlFactory(@"C:\Temp\ServiceControl");
-
-            // TODO: ServiceControl SQL Broker Connection String
-            var transport = new SqlTransportDetails("Data Source=.\\SQLEXPRESS;Initial Catalog=nservicebus;Integrated Security=True");
-
-            // TODO: Use transport to Clean Up
-
-            serviceControl = factory.Start(transport, context.Test.Name);
-        }
+            { typeof(SqlTransportDetails), () => new SqlTransportDetails("Data Source=.\\SQLEXPRESS;Initial Catalog=nservicebus;Integrated Security=True") }
+        };
 
         [TearDown]
         public void TearDown()
         {
-            serviceControl.Stop();
+            // TODO: Use transport to Clean Up
+
+            serviceControl?.Stop();
+        }
+
+        protected ITransportDetails ActivateInstanceOfTransportDetail(Type transportDetailType)
+        {
+            Func<ITransportDetails> activator;
+            transportDetailActivations.TryGetValue(transportDetailType, out activator);
+
+            return activator?.Invoke();
+        }
+
+        protected void StartServiceControl(ITransportDetails transport)
+        {
+            // TODO: This is a build artifact from the latest SC build. It should be imported into the working folder of the build
+            var factory = new ServiceControlFactory(@"C:\Temp\ServiceControl");
+
+            serviceControl = factory.Start(transport, TestContext.CurrentContext.Test.Name);
         }
 
         protected ServiceControlApi ServiceControl => serviceControl.Api;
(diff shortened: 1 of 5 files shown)


src/PlatformCompatibilityTests/ServiceControlCompatibilityTests/Infrastructure/Sql/SqlScTest.cs
```diff
@@ -1,33 +1,41 @@
...
+            return activator?.Invoke();
         }
 
-        [TearDown]
-        public void TearDown()
+        protected void StartServiceControl(ITransportDetails transport)
         {
-            serviceControl.Stop();
+            // TODO: This is a build artifact from the latest SC build. It should be imported into the working folder of the build
+            var factory = new ServiceControlFactory(@"C:\Temp\ServiceControl");
```
HEskandari: Is this a constant path? in which case move to config file?
but even better would be reading from Environment variables and the it will be just a config from TC to get it running.

src/PlatformCompatibilityTests/ServiceControlCompatibilityTests/Infrastructure/Sql/SqlScTest.cs
```diff
@@ -1,33 +1,41 @@
...
-            // TODO: ServiceControl SQL Broker Connection String
-            var transport = new SqlTransportDetails("Data Source=.\\SQLEXPRESS;Initial Catalog=nservicebus;Integrated Security=True");
+            serviceControl?.Stop();
+        }
 
-            // TODO: Use transport to Clean Up
+        protected ITransportDetails ActivateInstanceOfTransportDetail(Type transportDetailType)
+        {
+            Func<ITransportDetails> activator;
+            transportDetailActivations.TryGetValue(transportDetailType, out activator);
```
HEskandari: Why TryGetValue? Wouldn't there always be a key for the configured datasource? Should we Assert.Fail if the type is not configured?

src/PlatformCompatibilityTests/ServiceControlCompatibilityTests/SampleTests.cs
```diff
@@ -1,26 +1,50 @@
...
+    using System.Linq;
 
     [TestFixture]
     class SampleTests : SqlScTest
     {
         [Test]
-        public void SampleTestA()
+        public async void SampleTestA()
         {
-            var details = ServiceControl.Get<ServiceControlDetails>("");
```
HEskandari: Rather than exposing "Get" method and having DTOs all over the place, think we''d be better off if ServiceControl wrapper would only expose a few meaningful methods to work with rather than generic ones. Might be out of scope for this PR though?

src/PlatformCompatibilityTests/ServiceControlCompatibilityTests/Infrastructure/Sql/SqlScTest.cs
mikeminutillo: Well the intent is that it gets copied in from the linked SC build in which case we can make it a relative path?

Environment Variable from TC is also a pretty good option. The problem will always be that you don't have this content when you clone the repo so it has to tell you what to do when it runs and fails. As an additional challenge, if you manually download a version onto your machine and set up an env variable to point to it, we have no way of knowing how old/out of date it is when you run the tests.

src/PlatformCompatibilityTests/ServiceControlCompatibilityTests/SampleTests.cs
mikeminutillo: Yeah this was copy-pasted from the SC tests with the intent that we would build something more use case specific here. The SC tests have access to all of the internal objects used for the API so the generic approach is easy there.

mikeminutillo:
I like this and I'm going to merge it so I can continue to build on top of it. 

This means that if we want to include Multi-Instance then we'll need a new `ITransportDetails` implementation to provide it. I think that's OK. It does mean that this is no longer a SqlScTest base class. It can be just an ScTest base class.

I wonder if we can push the actual test into the base class and then have a single test per class that is called from this base test.

src/PlatformCompatibilityTests/ServiceControlCompatibilityTests/Infrastructure/Sql/SqlScTest.cs
WilliamBZA: > The problem will always be that you don't have this content when you clone the repo so it has to tell you what to do when it runs and fails

Which is exactly one of the problems I had. so :+1: for downloading from TC if it doesn't exist or an update is available.

src/PlatformCompatibilityTests/ServiceControlCompatibilityTests/Infrastructure/Sql/SqlScTest.cs
WilliamBZA: Good call.

PR #148, opened 2016-06-22
Added test to verify that retries are processed successfully

WilliamBZA:
Connects to: https://github.com/Particular/V6Launch/issues/79

An initial take on the Successful Retries test.

Process follows:

* Start 2 endpoints (EndpointA and EndpointB)
* Make sure EndpointB constantly fails
* Send a message from EndpointA to EndpointB
* Wait until ServiceControl shows a new error message
* Change EndpointB to succeed
* Retry the message using ServiceControl POST endpoint
* Wait for 60 seconds or until EndpointB successfully processes the retry

Failure is determined by not processing the retry within the specified timeframe. The use of a timeframe could make this test a bit flaky, but I can't think of another way to detect a failure. I'm pretty certain there _has_ to be a time dimension here.

Ping @mikeminutillo @HEskandari @ramonsmits @gbiellem


src/PlatformCompatibilityTests/ServiceControlCompatibilityTests/Infrastructure/ServiceControlApi.cs
```diff
@@ -20,12 +23,62 @@ public bool CheckIsAvailable()
...
+            // Perhaps we should use a message mutator to hardcode the message id we're looking for instead?
+            var startMessages = await GetErrorsForEndpoint<IList<FailedMessage>>(endpointName).ConfigureAwait(false);
+            var startCount = startMessages.Count;
+
+            IList<FailedMessage> currentMessages;
+            int newCount;
+
+            do
+            {
+                Thread.Sleep(200);
```
HEskandari: ummm Task.Delay?

src/PlatformCompatibilityTests/ServiceControlCompatibilityTests/Infrastructure/ServiceControlApi.cs
```diff
@@ -20,12 +23,62 @@ public bool CheckIsAvailable()
...
+                currentMessages = await GetErrorsForEndpoint<IList<FailedMessage>>(endpointName).ConfigureAwait(false);
+                newCount = currentMessages.Count;
+            } while (newCount <= startCount);
+
+            var newFailedMessage = currentMessages.FirstOrDefault(m => !startMessages.Select(sm => sm.Id).Contains(m.Id));
```
HEskandari: .Single(...)?

src/PlatformCompatibilityTests/ServiceControlCompatibilityTests/SuccessfulRetryTests.cs
```diff
@@ -0,0 +1,74 @@
...
+            endpointA.Send("EndpointB", new TestMessage());
+
+            var failedMessage = await ServiceControl.WaitForNewFailingMessages("EndpointA");
+
+            handlerInstance.ShouldFail = false;
+            await ServiceControl.RetryMessageId(failedMessage.Id);
+
+            var handled = handlerInstance.WaitForSuccessfulMessage().Wait(60000); // Is this a reasonable time to wait?
+
+            Assert.IsTrue(handled, "Did not process the retry successfully within the time limit");
```
HEskandari: Maybe also log the time we waited (60 secs) that might come in handy when a test fails and we trace the build/test run logs.

HEskandari:
@WilliamBZA RE retry/wait, could we do a retry, poll error messages until the message is back to the error queue OR the Endpoint B has successfully processed it? that way you don't have to have a time dimension.

src/PlatformCompatibilityTests/ServiceControlCompatibilityTests/SuccessfulRetryTests.cs
```diff
@@ -0,0 +1,74 @@
...
+            var transportDetails = StartUp(transportDetailType);
+            await RunTest(transportDetails);
+        }
+
+        async Task RunTest(ITransportDetails transportDetails)
+        {
+            var handlerInstance = new TestMessageHandler();
+            var builder = new ContainerBuilder();
+            builder.RegisterInstance(handlerInstance);
+            var container = builder.Build();
```
HEskandari: IMO this whole block about registering the handler instance seem to be unrelated to the test. Could it be moved out to somewhere else? Some sort of extension/helper or somehow done in the Endpoint creation.

src/PlatformCompatibilityTests/ServiceControlCompatibilityTests/Infrastructure/ServiceControlApi.cs
WilliamBZA: There may be multiple new messages if some other process also sends failed messages to SC. That's why I think we might be better off using a mutator to look for a specific message ID.

src/PlatformCompatibilityTests/ServiceControlCompatibilityTests/Infrastructure/ServiceControlApi.cs
WilliamBZA: :+1:

src/PlatformCompatibilityTests/ServiceControlCompatibilityTests/SuccessfulRetryTests.cs
WilliamBZA: :+1:

src/PlatformCompatibilityTests/ServiceControlCompatibilityTests/SuccessfulRetryTests.cs
WilliamBZA: I'm not sure. Let me see what I can do. This is the Arrange part of the AAA, and honestly I don't like the singleton nature here anyway, but couldn't come up with a nicer way to manage state within the handler reliably. I'll try work on it some more today.

src/PlatformCompatibilityTests/ServiceControlCompatibilityTests/Infrastructure/ServiceControlApi.cs
mikeminutillo: :+1: for grabbing the Id if we can. Not sure if we have an API to override the ID though.

mikeminutillo: Isn't this also a potential timing conflict too? If SC manages to ingest the message before we make our first API call we'll block forever.

src/PlatformCompatibilityTests/ServiceControlCompatibilityTests/SuccessfulRetryTests.cs
HEskandari: the arrange could be just the creation of the handler + endpoint, registration of the container, etc. is just noise and is not directly related to the test logic (why should you have to change the test codes if you use another container).

RE the state, would it be possible to send the flag via the Context to the handler?

src/PlatformCompatibilityTests/ServiceControlCompatibilityTests/Infrastructure/ServiceControlApi.cs
HEskandari: Could we leverage [Timeout] for that to ensure we're not blocking indefinitely?

WilliamBZA: Yes it is. This is super flaky as it stands now.

I'll try get a hardcoded MessageID in place.

src/PlatformCompatibilityTests/ServiceControlCompatibilityTests/SuccessfulRetryTests.cs
mikeminutillo: I have an idea for this but my VS has been taken out of commission by an incredibly slow update today.

Something like this?

```
var endpointDetails = new EndpointDetails(new TestContext { ShouldFail = true });
endpointDetails.Add<TestMessageHandler>();

var endpointA = await CreateSqlEndpoint("EndpointA", transportDetails);
var endpointB = await CreateSqlEndpoint("EndpointB", transportDetails, endpointDetails);

endpointA.Send(endpointB, new TestMessage());

var failedMessage = await ServiceControl.WaitForNewFailingMessages(endpointA);

endpointDetails.Context.ShouldFail = false;

await ServiceControl.RetryMessageId(failedMessage.Id);

var handled = endpointDetails.Context.WaitForDone().Wait(60000);

Assert.IsTrue(handled, "blah blah");



class TestContext
{
  TaskCompletionSource<bool> doneSource = new TaskCompletionSource<bool>();

  public bool ShouldFail { get; set; }
  
  public void MarkSuccess()
  {
    doneSource.SetResult(true);
  }

  public Task WaitForDone()
  {
    return doneSource.Task;
  }
}

class TestMessageHandler : IHandleMessages<TestMessage>
{
  public TestMessageHandler(TestContext testContext)
  {
    this.testContext = testContext;
  }

  public Task Handle(TestMessage message, IMessageHandlerContext context)
  {
    if(testContext.ShouldFail)
    {
      throw new ApplicationException("Failing...");
    }

    testContext.MarkSuccess();
    return Task.FromResult(0);
  }

  private TestContext testContext;
}

class EndpointDetails<TContext>
{
  public EndpointDetails(TContext context)
  {
    Context = context;
  }

  public void Add<T>()
  {
    // capture T in a list somewhere
  }

  public IContainer CreateContainer()
  {
    var builder = new ContainerBuilder();
    builder.RegisterInstance(Context);
    // For each of the items added above by `Add`, add them to the container
    return builder.Build();
  }

  public TContext Context { get; }
  
}
```

WilliamBZA: wow

mikeminutillo:
OK, I did some refactoring to introduce some new concepts.
There is a TestContextBase that can be injected into endpoints so that handlers can signify the end of the test.

There is an EndpointDetails class that is just a wrapper around constructing the container. It lets us specify the classes that get loaded into an endpoint fairly succinctly. 

I added an EndpointFactory that hides away the Transport Details from the test. As long as it has that it can spin up the endpoints (we don't stop these anywhere right?)

I modified the Send so that we can pass in a messageId and then add a WaitForFailedMessage that can wait for a specific message to show up. 

It all seems to come together fairly cleanly.

I'll try and add the Reply Handler Test using this infrastructure and see what it looks like.

mikeminutillo:
Yep. That works :)

mikeminutillo:
Made some quick changes to push the name into the EndpointDetails and return the messageId from Send (rather than specify it).

PR #149, opened 2016-06-26
Include rabbitmq, MSMQ, ASQ and ASB

WilliamBZA:
Connects to: https://github.com/Particular/V6Launch/issues/79

Dependent on: https://github.com/Particular/EndToEnd/pull/148

Adds MSMQ and RabbitMQ transports to the test categories.

Ping @HEskandari @mikeminutillo @ramonsmits @gbiellem


WilliamBZA:
Added the other transports too

src/PlatformCompatibilityTests/Sql/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,36 @@
...
+//      Major Version
+//      Minor Version 
+//      Build Number
+//      Revision
+//
+// You can specify all the values or you can default the Build and Revision Numbers 
+// by using the '*' as shown below:
+// [assembly: AssemblyVersion("1.0.*")]
+[assembly: AssemblyVersion("1.0.0.0")]
+[assembly: AssemblyFileVersion("1.0.0.0")]
```
HEskandari: Errr....I don't think this is the right AssemblyInfo. Probably need to copy one from other projects with the right Copyright/Company/etc.

HEskandari:
Just a side note that I don't see TeamCity build status on the PRs. Something we need to setup like we did for NHibernate Compat testing maybe.

src/PlatformCompatibilityTests/ASQ/ASQTransportDetails.cs
```diff
@@ -0,0 +1,35 @@
...
+
+namespace ServiceControlCompatibilityTests
+{
+    public class ASQTransportDetails : ITransportDetails
+    {
+        const string TransportTypeName = "NServiceBus.AzureStorageQueueTransport, NServiceBus.Azure.Transports.WindowsAzureStorageQueues";
+
+        public ASQTransportDetails(string connectionString)
+        {
+            connectionString = "UseDevelopmentStorage=true";
```
mikeminutillo: Should this be a `+=` or something? Why are we overwriting what gets passed in?

src/PlatformCompatibilityTests/ServiceControlCompatibilityTests/Infrastructure/Msmq/MsmqTransportDetails.cs
```diff
@@ -0,0 +1,25 @@
...
+
+        public void ApplyTo(Configuration configuration)
+        {
+            var settings = configuration.AppSettings.Settings;
+            settings.Set(SettingsList.TransportType, TransportTypeName);
+        }
+
+        public void ConfigureEndpoint(EndpointConfiguration endpointConfig)
+        {
+            endpointConfig.PurgeOnStartup(true);
```
mikeminutillo: Should we rely on the defaults here or explicitly set the transport to MSMQ?

mikeminutillo:
Overall it looks good and supporting those other transports is surprisingly painless.

src/PlatformCompatibilityTests/ASQ/ASQTransportDetails.cs
WilliamBZA: Uhh oops, that shouldn't have made it in...

src/PlatformCompatibilityTests/ServiceControlCompatibilityTests/Infrastructure/Msmq/MsmqTransportDetails.cs
WilliamBZA: Defaults should be fine. We could be explicit about it for clarity. Don't think there'd be much more benefit than that.

PR #157, opened 2016-07-04
Multi-Instance SQL Transport support for platform compatibility tests

mikeminutillo:
Each test may now provide a mapping from endpoint name to some kind of tag (in this case I've just used logical server name). It is up to the transport to map that to something useful if it wants to.

Multi-Instance SQL Transport uses the tag to map each endpoint to a specific connection string.


WilliamBZA:
LGTM :+1:

mikeminutillo:
Alright, I am going to merge it down so that it runs on the build server

Issue #160, opened 2016-07-27
Perfests: Consider using Stopwatch instead of DateTime in receive statistics


ramonsmits:
@bording suggested to use Stopwatch instead of DateTime as Stopwatch should be more lightweight thus should result in an improved throughput.

The relative performance between V5 and V6 would remain the same but absolute performance should increase for both.

dvdstelt:
@ramon did you see any differences in perf?

PR #166, opened 2016-07-29
Implemented #160, statistics now using Stopwatch. 

ramonsmits:
Refactoring suggested by #160 

Access to statistics also made faster by removing step that also did a null check at each invocation.

The changes look more efficient but I could not measure a repeatable improvement.


ramonsmits:
@dvdstelt To answer your question on the issue. I measured but in the grand scheme of things a `DateTime.UtcNow` vs `Stopwatch.GetTimestamp` is (luckily) not really measurable.

PR #175, opened 2016-08-23
Upgrade transports compat tests to beta 7

weralabaj:


ramonsmits:
@weralabaj  What was the reason that it previously didn't work?

src/TransportCompatibilityTests/AzureServiceBusV7/EndpointFacade.cs
```diff
@@ -27,16 +25,15 @@ public class EndpointFacade : MarshalByRefObject, IEndpointFacade
...
                 .ConnectionString(AzureServiceBusConnectionStringBuilder.Build)
                 .Sanitization().UseStrategy<ValidateAndHashIfNeeded>();
 
             // TODO: remove when core v6 & asb V7 package are updated
             endpointConfiguration.UseSerialization<JsonSerializer>(); 
-            // Needed for callbacks package to work
-            endpointConfiguration.ScaleOut().InstanceDiscriminator(Guid.NewGuid().ToString("N"));
 
             endpointConfiguration.CustomConfigurationSource(new CustomConfiguration(endpointDefinition.Mappings));
+            endpointConfiguration.MakeInstanceUniquelyAddressable(Guid.NewGuid() + "N");
```
ramonsmits: Why not "A" here like for Rabbit and SQL and why the need to append the "N"?

weralabaj: Not sure, I just copied what was there before :P I'll change it.

weralabaj:
> @weralabaj What was the reason that it previously didn't work?

Corrupted csproj file :P

PR #185, opened 2016-08-31
Make serializer compat tests work again

andreasohlund:


andreasohlund:
Seems like this has been broken for a while, all tests fails with

`OneTimeSetUp: SetUpAttribute attribute not allowed in a SetUpFixture` nunit 3 issue?

andreasohlund:
@Particular/endtoend-maintainers ideas? seems like we need to stop using the OneTimeSetUp on the base fixture but that screwed up (I think) the cleanup. Perhaps separate base fixtures for the different tests "categories"?

andreasohlund:
ping @Particular/endtoend-maintainers I need some help with this one. Would anyone be able to pair up?

danielmarbach:
@andreasohlund made them work again except for the GenericMessage tests

danielmarbach:
Green

danielmarbach:
When this works we can add tests for the DateTime compatibility and potentially start fixing the generic message test

PR #189, opened 2016-09-02
Setting different Categories for Test Fixtures 

WojcikMike:
to enable running tests in smaller groups on builder

This step would allow to create separate build configuration running only subset of tests like:
- Transports
- Permutations 
  etc.

By doing that we would allow to run only subset of test: only Transports. And after each step we would get full feedback with artifacts etc.

cc: @Particular/endtoend-maintainers


WojcikMike:
ping @Particular/endtoend-maintainers can you have a look?

danielmarbach:
@WojcikMike seems you never got a response. Since the PR is outdated now I'm closing this. @ramonsmits feel free to raise this again on latest develop if you want this.

PR #192, opened 2016-09-05
Introducing async countdown into loop runner and fixing shutdown proxy wait

danielmarbach:
Removes the TaskFactory usage including the `LongRunning` flag. The long running would not have positive impact on this code. After the first await statement the dedicated background thread would potentially already be gone. Switched to `Task.Run` instead.

Inner send loop was a mixture of Async and Sync blocking calls. Introduced AsyncCountDown event which makes it possible to asynchronously await a batch of messages.

### DIFF
--- a/src/PerformanceTests/Utils/AsyncCountdownEvent.cs
+++ b/src/PerformanceTests/Utils/AsyncCountdownEvent.cs
@@ -1,0 +1,106 @@
+﻿using System;
+using System.Threading;
+using System.Threading.Tasks;
+
+/// <remarks>
+/// Based on https://github.com/StephenCleary/AsyncEx
+/// </remarks>
+public sealed class AsyncCountdownEvent
+{
+    public AsyncCountdownEvent(int count)
+    {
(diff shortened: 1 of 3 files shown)


danielmarbach:
@Particular/endtoend-maintainers please review and merge

src/PerformanceTests/Utils/AsyncCountdownEvent.cs
```diff
@@ -0,0 +1,103 @@
+﻿using System;
+using System.Threading;
+using System.Threading.Tasks;
+
+public sealed class AsyncCountdownEvent
```
dvdstelt: There are quite a few variations of this class floating the internet. Would you care providing a source so we can see if there's an update, if we run into trouble or anything?

danielmarbach: It is a mixture of https://github.com/StephenCleary/AsyncEx in combination with my own code.

danielmarbach:
@ramonsmits @dvdstelt can you guys help me rebase this?

PR #195, opened 2016-09-07
Loop runner now using Parallel.For to prevent sequential processing if persistence implementation isn't async.

### LINKED ISSUE #98: TaskHelper runs sequentially for V5 due to Task.FromResult(0) in V5 Session implementation
TaskHelper runs sequentially for V5 due to Task.FromResult(0) in V5 Session implementation although its expected that all tasks are processed concurrently.

## Code

* https://github.com/Particular/EndToEnd/blob/perf-tests/src/PerformanceTests/Common/Scenarios/LoopRunner.cs#L60

```c#
 await TaskHelper.ParallelFor(BatchSize, () => SendMessage(session));
```

* https://github.com/Particular/EndToEnd/blob/perf-tests/src/PerformanceTests/Utils/TaskHelper.cs#L6

```c#
var sends = new Task[count];
for (var i = 0; i < count; i++) sends[i] = action();
await Task.WhenAll(sends);
```

* https://github.com/Particular/EndToEnd/blob/perf-tests/src/PerformanceTests/NServiceBus5/Session.cs#L36

```c#
if (localAddress == null)
    instance.SendLocal(message);
else
    instance.Send(localAddress, message);

return Task.FromResult(0);
```

## Possible solutions:

- Convert to a Parallel.For
- Use Task.Yield to break sequential processing

Thanks @SzymonPobiega for this catch!

ramonsmits:
Resolves #98

### DIFF
--- a/src/PerformanceTests/Common/Scenarios/LoopRunner.cs
+++ b/src/PerformanceTests/Common/Scenarios/LoopRunner.cs
@@ -52,6 +52,8 @@ abstract class LoopRunner : BaseRunner
 
             var cancellationToken = stopLoop.Token;
 
+            var po = new ParallelOptions { MaxDegreeOfParallelism = Environment.ProcessorCount };
+
             while (!Shutdown)
             {
                 try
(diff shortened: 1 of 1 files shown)


ramonsmits:
@dvdstelt Please squash and not merge.

PR #197, opened 2016-09-07
Debug build now has Debug log output

ramonsmits:
Previously the debug build ran with Info log output. Debug builds are usually ran for resolving issues and debug output as default prevents the need to adjust the log level (developers are lazy...).


ramonsmits:
@dvdstelt Please squash and not merge.

Issue #198, opened 2016-09-08
Test case that does saga updates


ramonsmits:
We currently have the _saga initiate_ scenario which seeds a lot of message and result in each creating a new saga instance. All messages can be processed concurrently as there cannot be any contention.

In issue https://github.com/Particular/PlatformDevelopment/issues/226 its requested that we also want to know what the performance is for sagas that are updated without contention.

Current idea is for a case is:
- Look at max concurrency level
- Seed that many messages that create a saga instance
- Saga receives the messages, does an update (increase value) and sends a message to itself
- Repeat

This will _never_ have contention on the same saga instance and would give us a number on simple saga update performance.

ramonsmits:
Implemented https://github.com/Particular/EndToEnd/tree/master/src/PerformanceTests/Scenarios/SagaUpdate

PR #200, opened 2016-09-08
Saga update runner

ramonsmits:
Implements #198

### DIFF
--- a/src/PerformanceTests/Common/Scenarios/BaseRunner.cs
+++ b/src/PerformanceTests/Common/Scenarios/BaseRunner.cs
@@ -89,7 +89,7 @@ public abstract class BaseRunner : IConfigurationSource, IContext
 
     protected virtual Task Stop()
     {
-        return Task.FromResult(0);
+        return DrainMessages();
     }
 
     async Task CreateSeedData(ICreateSeedData instance)
--- a/src/PerformanceTests/Common/Scenarios/BaseRunner.cs
+++ b/src/PerformanceTests/Common/Scenarios/BaseRunner.cs
@@ -29,6 +29,9 @@ public abstract class BaseRunner : IConfigurationSource, IContext
 
     protected byte[] Data { private set; get; }
     protected bool SendOnly { get; set; }
+    protected int MaxConcurrencyLevel { private set; get; }
+
+    protected static bool Shutdown { private set; get; }
 
     public async Task Execute(Permutation permutation, string endpointName)
     {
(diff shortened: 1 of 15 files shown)


src/PerformanceTests/Common/Scenarios/SagaUpdate/SagaUpdateRunner.cs
```diff
@@ -0,0 +1,52 @@
...
+partial class SagaUpdateRunner
+    : BaseRunner
+{
+    public static bool Shutdown;
+    readonly ILog Log = LogManager.GetLogger(nameof(SagaUpdateRunner));
+
+    protected override async Task Start(ISession session)
+    {
+        var maxConcurrencyLevel = ConcurrencyLevelConverter.Convert(Permutation.ConcurrencyLevel);
+        var seedSize = maxConcurrencyLevel * 2;
```
dvdstelt: I've seen this quite a number of times. Isn't this something you could do in BaseRunner or some helper class perhaps?

dvdstelt:
Rest seems legit, a lot of the same stuff going on. Do it once, okay. Do it twice, good as well. Do it three times, you should probably refactor. How many times is code like this in the repo?
On the other hand, I don't think it needs to become more complex by adding yet another base class or something.

What do you think?

src/PerformanceTests/Common/Scenarios/SagaUpdate/SagaUpdateRunner.cs
ramonsmits: Have been considering to but the calculated concurrency value as a protected property in the base class. Not a real saver but now not each scenario needs to be responsible for knowing how to calculate it.

Same for the logic that drains the queue although I have been thinking about creating a behavior for that to very early in the pipeline.

src/PerformanceTests/Common/Scenarios/BaseRunner.cs
```diff
@@ -302,4 +307,21 @@ static IEnumerable<int> IterateUntilFalse(Func<bool> condition)
         var i = 0;
         while (condition()) yield return i++;
     }
+
+    protected async Task DrainMessages()
```
dvdstelt: I think this method is called in the `Close()` method of every runner. If this is the case, isn't it better to add it to `Stop()` method in the `BaseRunner` until someone actually does NOT need it and another approach can be taken?

ramonsmits: @dvdstelt check the latest commit

PR #231, opened 2016-10-03
WIP: Improvements

ramonsmits:
CSV/TSB report generator 

- not overwriting existing report by using timestamp in filename
- NA values but just empty cells
- Logging now writing individual permutation property values to these show up as columns

Azure Service Bus transport

- Suppressing ObjectDisposedException failures to workaround V7 bug

Seeding

- Default seed/receive ratio lowered as seeding has become faster on all transports
- Formatting of seed window expired message
- Adaptive SeedDurationFactor using history of previous permutation runs to run a seed/receive test as close to the wanted test duration

Processing

- Detect if there are still incoming message to stop the test earlier instead of just waiting until the run duration is elapsed

Purging/draining

- CreateOrPurge now drains messages from the queue
- Message draining shortcuts the pipeline to drain ASAP

Utility

- Added TaskWhenAllThrottled for dev test purposes

### DIFF
--- a/src/PerformanceTests/Common/Scenarios/BaseRunner.cs
+++ b/src/PerformanceTests/Common/Scenarios/BaseRunner.cs
@@ -14,6 +14,8 @@ using NServiceBus.Logging;
 using Tests.Permutations;
 using System.Collections.Generic;
 using System.Diagnostics;
+using System.Globalization;
+using System.IO;
 using System.Threading;
 using System.Threading.Tasks;
 using Common.Scenarios;
@@ -22,6 +24,9 @@ using Variables;
 public abstract class BaseRunner : IConfigurationSource, IContext
 {
     readonly ILog Log = LogManager.GetLogger("BaseRunner");
+
+    string SeedReceiveRatioPath => Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.ApplicationData), "Perftests", Permutation.Category, Permutation.Description, Permutation.Tests[0], Permutation.Id, "SeedDurationFactor.txt");
+    double seedAvg;
 
     public Permutation Permutation { get; private set; }
     public string EndpointName { get; private set; }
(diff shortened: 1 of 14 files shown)


src/PerformanceTests/Common/Scenarios/BaseRunner.cs
```diff
@@ -354,18 +449,51 @@ static IEnumerable<int> IterateUntilFalse(Func<bool> condition)
...
+            await Task.Delay(DrainPollInterval).ConfigureAwait(false);
+        } while (NServiceBus.Performance.StatisticsBehavior.ShortcutCount > current);
 
         var diff = current - startCount;
-        Log.InfoFormat("Drained {0} message(s)", diff);
+        Log.InfoFormat("Drained {0:N0} message(s) in {0:N0}ms", diff, start.ElapsedMilliseconds);
+        NServiceBus.Performance.StatisticsBehavior.Shortcut = false;
+    }
+
+    void ReadPermutationSeedDurationFactor()
```
dvdstelt: These two methods can probably be refactored into their own class?
These methods also aren't clear to me after 3 seconds of reading. Perhaps make them a bit more self explanatory?

src/PerformanceTests/Common/Scenarios/BaseRunner.cs
```diff
@@ -147,12 +239,14 @@ protected virtual Task Stop()
     }
 
 #if Version5
-    Task CreateOrPurgeQueues()
+    async Task CreateOrPurgeQueues()
     {
         var configuration = CreateConfiguration();
         if (IsPurgingSupported) configuration.PurgeOnStartup(true);
```
ramonsmits: @dennis Some transports do support purging (MSMQ, SQL), some support async purging (ASQ) some don't at all. This makes sure that the incoming queue is empty no matter what the support for purging is.

src/PerformanceTests/Common/Scenarios/BaseRunner.cs
ramonsmits: @dvdstelt Yes, they should be extracted as it wont even work on Azure due to VM always being created. The idea here is that the seed duration is adaptive per permutation. The more often you run it the better the seed/receive throughput aligns with the specified test duration.

I'm also not sure if I want this on the master branch yet.

ramonsmits:
Obsolete due to individual PR's

src/PerformanceTests/Common/Scenarios/BaseRunner.cs
dvdstelt: I get why they are drained this way. What I don't get is why they are always drained? As far as I can remember, this was optional to drain queues?

ramonsmits: Not draining can result in poisenous messages from other tests to conflict with the new test.

PR #242, opened 2016-10-12
Front-end for Performance Tests

ramonsmits:
A simple front-end to select any permutation value and then launch that specific permutation.

image

It's ugly, it's created fast, but it works :-)

Needed a little refactoring on the TestEnvironment class which is reused.

Also not proud on the argument parsing for run/warmup durations.


danielmarbach:
Is this still required?

ramonsmits:
This would require significant updates which are currently not worth the effort.

PR #248, opened 2016-10-18
Resolves bug related to draining messages at start

ramonsmits:
Draining during purge operation didn't work properly as messages got processed before the draining could be started due to messages already being received and processed before the endpoint is completely initialized when calling `await Endpoint.Start(configuration)`

### DIFF
--- a/src/PerformanceTests/Common/Scenarios/BaseRunner.cs
+++ b/src/PerformanceTests/Common/Scenarios/BaseRunner.cs
@@ -43,7 +43,8 @@ public abstract class BaseRunner : IConfigurationSource, IContext
         InitData();
         MaxConcurrencyLevel = ConcurrencyLevelConverter.Convert(Permutation.ConcurrencyLevel);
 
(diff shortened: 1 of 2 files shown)


src/PerformanceTests/Common/Scenarios/BaseRunner.cs
```diff
@@ -91,8 +92,6 @@ protected virtual Task Stop()
 
     async Task CreateSeedData(ICreateSeedData instance)
     {
-        Log.Info("Creating or purging queues...");
-        await CreateOrPurgeAndDrainQueues().ConfigureAwait(false);
         Log.Info("Creating send only endpoint...");
```
dvdstelt: Can't this be 4 dots?

ramonsmits: Well, even better is replacing this with the [horizontal ellipsis character (…)](http://www.fileformat.info/info/unicode/char/2026/index.htm).

What do you think?

dvdstelt: Ah, now I get it... Thanks for clarifying @ramonsmits. I think three dots will do just fine 😸

PR #258, opened 2017-10-05
NServiceBus V7 Support

ramonsmits:
## Included packages

- [x] Windows performance counters
- [x] Transport: MSMQ
- [x] Transport: SQL
- [x] Transport: SQS
- [x] Persistence: In memory
- [x] Persistence: RavenDB
- [x] Persistence: NHibernate?
- [x] Serializer: Newtonsoft.Json
- [x] Transport: RabbitMQ
- [x] Transport: ASQ
- [x] Transport: ASB
- [x] Persistence: Azure
- [x] Transport: SQS (Core 6)


ramonsmits:
@andreasohlund @timbussmann WIP of v7 host

timbussmann:
I get tons of compilation errors due to obsolete API usage when I try to build this locally.

ramonsmits:
@timbussmann Try this version to compile.

ramonsmits:
Currently, the sagas are part of the common project which makes it hard to add support for the SQL persister because it is the only persister that uses its own saga base class.

Potentially it might work by moving the sagas to a common saga project shared among persisters. In the common project the sagas could be partials where the persisters would add their partial parts to actually be able to compile the sagas or use a `#define` or something.

@SimonCropp @dvdstelt @timbussmann @andreasohlund any ideas on how to add support for SQL persistence and still share the sagas code? Do we have something similar with AT's for persisters?

src/PerformanceTests/Tests/Categories/AzureFixture.cs
```diff
@@ -26,10 +18,13 @@ static IEnumerable<Permutation> CreatePermutations()
         {
             return PermutationGenerator.Generate(new Permutations
             {
-                Transports = new[] { Transport.AzureStorageQueues, Transport.AzureServiceBus,  },
-                Persisters = new [] { Persistence.Azure, },
-                Serializers = new[] { Serialization.Json, },
-                OutboxModes = new[] { Outbox.Off, },
+                Versions = new[] { NServiceBusVersion.V7 },
+                Transports = new[] { Transport.MSMQ },
```
timbussmann: why MSMQ?

src/PerformanceTests/Common/Scenarios/BaseRunner.cs
```diff
@@ -98,19 +96,28 @@ protected virtual Task Stop()
 
     async Task CreateSeedData(ICreateSeedData instance)
     {
-        Log.Info("Creating send only endpoint...");
-        await CreateSendOnlyEndpoint().ConfigureAwait(false);
+        var sendOnly = IsSendOnly;
+        IsSendOnly = true; // Needed as performance counter profiles apply configuration that is not allowed on send-only endpoints.
```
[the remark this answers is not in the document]
ramonsmits: @timbussmann See the comment :-)
>Needed as performance counter profiles apply configuration that is not allowed on send-only endpoints.

This is required only during seeding. Seeding is always SendOnly.

src/PerformanceTests/Tests/Categories/AzureFixture.cs
ramonsmits: @timbussmann Because MSMQ is the least impacting any persistence performance-related tests.

ramonsmits:
@dvdstelt Made changes based on the review comments. 

@timbussmann Would you like to see some changes based on your comments?

src/PerformanceTests/Tests/Categories/AzureFixture.cs
andreasohlund: Looking at the acceptance tests it definitely seems like the learning transport is faster than MSMQ, would be interesting to see if that would hold true for the perftests as well?

src/PerformanceTests/Common/Scenarios/BaseRunner.cs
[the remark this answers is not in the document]
timbussmann: I still don't understand the code. It states something about not working on send-only, but it sets IsSendOnly to true. I also don't understand why we need to enable a feature which doesn't work on send only endpoints.

PR #261, opened 2017-10-11
SQL Persistence custom saga strategy

ramonsmits:
Potential strategy to have persistence specific overrides for sagas by moving saga code into a source code project, sharing it among persistence projects and having the ability to add custom code via partials.

- One saga is moved from Common to a new project called 'Scenarios'.
- This Scenarios project would need to be included into every persistence profile project in order for it to have the code to execute the tests.
- Custom saga code can be added by having the persistence profile project have a `CustomSaga` flag set.

Please share your opinion about this solution strategy and mention potentially any alternatives. Focus here is to prevent recompiles hence moving the persistence specific code instead of having a persistence specific implementation in the common project that can be selected via a compiler flag.

https://github.com/Particular/DotNetCoreLaunch/issues/54


SzymonPobiega:
So the approach is to use the `customsaga` compilation flag to determine if the saga is included in the base class. Otherwise (SQL persistence) it is defined in the actual test. Do I understand correctly?

ramonsmits:
@SzymonPobiega The sagas are split into several partials. Basically one partial that has the saga data, and handlers and another that depends on if it 'custom' or not.


The shared source code for a saga for all persisters:

- https://github.com/Particular/EndToEnd/blob/sqlpersistence-v7/src/PerformanceTests/Scenarios/SagaInitiate/SagaInitiateRunner.V6.cs#L8-L21

The shared source code for a saga for all persisters that do not have a custom base class:
- https://github.com/Particular/EndToEnd/blob/sqlpersistence-v7/src/PerformanceTests/Scenarios/SagaInitiate/SagaInitiateRunner.V6.cs#L23-L30

The source code for the SQL persister that is persister specific:
- https://github.com/Particular/EndToEnd/blob/sqlpersistence-v7/src/PerformanceTests/Persistence.V7.SQL.V3/SagaInitiate/SagaInitiateRunner.V6.cs#L6-L15

It is currently all shared code as this cannot be done with types

weralabaj:
I've reviewed the code and also did some thinking over the weekend, I don't have a better idea of how this could be tackled. Additionally, given that the code is working, I don't see any reasons to look for another approach now :)
I don't think what we do with Acceptance tests in SQL persistence repo is relevant here, it's a different scenario (simply overwriting some of the tests).

PR #266, opened 2017-10-20
Fix v7 host by adding ConcurrencyProfile

bording:
@ramonsmits This is the reason for the major performance regression I noticed in the v7 test results. Looks like the max concurrency wasn't being set properly with the v7 host runner.


ramonsmits:
@bording Did you manage to run the RabbitMQ tests too?

PR #270, opened 2017-11-03
Remove perf counters

timbussmann:
To be able to target .net standard, we need to replace our usage of performance counters.

There seem to be other, process relates data about memory usage we might use.
Thoughts @ramonsmits @bording 

(this branch needs to be retargeted once csproj-update has been merged)

### DIFF
--- a/src/PerformanceTests/Utils/Statistics.cs
+++ b/src/PerformanceTests/Utils/Statistics.cs
@@ -1,5 +1,4 @@
 ﻿using System;
-using System.Collections.Concurrent;
 using System.Diagnostics;
 using System.Linq;
 using System.Threading;
@@ -22,10 +21,6 @@ public class Statistics : IDisposable
     TimeSpan sendTimeWithTx = TimeSpan.Zero;
 
     Process process;
-    PerformanceCounter privateBytesCounter;
-    Timer perfCountersTimer;
-
-    static ConcurrentBag<double> perfCounterValues = new ConcurrentBag<double>();
 
     static ILog logger = LogManager.GetLogger("Statistics");
 
@@ -48,8 +43,6 @@ public class Statistics : IDisposable
         if (disposing)
         {
             process?.Dispose();
-            privateBytesCounter?.Dispose();
-            perfCountersTimer?.Dispose();
             GC.SuppressFinalize(this);
         }
     }
@@ -74,12 +67,6 @@ public class Statistics : IDisposable
     Statistics()
     {
         process = Process.GetCurrentProcess();
-        privateBytesCounter = new PerformanceCounter("Process", "Private Bytes", process.ProcessName);
-        perfCountersTimer = new Timer(state =>
-            perfCounterValues.Add(privateBytesCounter.NextValue()),
-            null,
-            TimeSpan.FromSeconds(1),
-            TimeSpan.FromSeconds(1));
     }
 
     static long GetTimestamp()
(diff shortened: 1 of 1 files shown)


src/PerformanceTests/Utils/Statistics.cs
```diff
@@ -131,10 +117,10 @@ public void Dump()
         if (sendTimeWithTx != TimeSpan.Zero)
             LogStats("SendingInsideTX", Convert.ToDouble(NumberOfMessages / 2) / sendTimeWithTx.TotalSeconds, "msg/s", StatsFormatDouble);
 
-        var counterValues = perfCounterValues.ToList();
-        LogStats("PrivateBytes-Min", counterValues.Min() / 1024, "kb", StatsFormatInt);
-        LogStats("PrivateBytes-Max", counterValues.Max() / 1024, "kb", StatsFormatInt);
-        LogStats("PrivateBytes-Avg", counterValues.Average() / 1024, "kb", StatsFormatInt);
+        LogStats("PrivateMemorySize", process.PrivateMemorySize64 / 1024, "kB", StatsFormatInt);
+        LogStats("PeakWorkingSet", process.PeakWorkingSet64 / 1024, "kB", StatsFormatInt);
+        LogStats("PeakPagedMemorySize", process.PeakPagedMemorySize64 / 1024, "kB", StatsFormatInt);
```
ramonsmits: > kB

src/PerformanceTests/Utils/Statistics.cs
```diff
@@ -131,10 +117,10 @@ public void Dump()
         if (sendTimeWithTx != TimeSpan.Zero)
             LogStats("SendingInsideTX", Convert.ToDouble(NumberOfMessages / 2) / sendTimeWithTx.TotalSeconds, "msg/s", StatsFormatDouble);
 
-        var counterValues = perfCounterValues.ToList();
-        LogStats("PrivateBytes-Min", counterValues.Min() / 1024, "kb", StatsFormatInt);
-        LogStats("PrivateBytes-Max", counterValues.Max() / 1024, "kb", StatsFormatInt);
-        LogStats("PrivateBytes-Avg", counterValues.Average() / 1024, "kb", StatsFormatInt);
+        LogStats("PrivateMemorySize", process.PrivateMemorySize64 / 1024, "kB", StatsFormatInt);
+        LogStats("PeakWorkingSet", process.PeakWorkingSet64 / 1024, "kB", StatsFormatInt);
```
ramonsmits: Wow... those were actually present on the process info? Freaking awesome!

src/PerformanceTests/Utils/Statistics.cs
```diff
@@ -76,12 +69,6 @@ public static IDisposable Initialize(Permutation permutation)
     {
         this.permutation = permutation;
         process = Process.GetCurrentProcess();
-        privateBytesCounter = new PerformanceCounter("Process", "Private Bytes", process.ProcessName);
-        perfCountersTimer = new Timer(state =>
```
ramonsmits: Why removed the timer? Would that be replaced with a call to `process.PrivateMemorySize64`?

timbussmann: not sure I understand the comment. Do you suggest to keep the timer to calculate an average of `process.PrivateMemorySize64`?

ramonsmits: That was the existing behavior. I don't know of how much use it is. @dvdstelt What do you think? Shall we just drop this? The peak memory usage is probably already a good indicator as most tests would just run 1 minute.

timbussmann:
rebased on master

PR #271, opened 2017-11-06
Switch logging to log4net

timbussmann:
Switch from nlog to log4net as log4net supportes netstandard

### DIFF
--- a/src/PerformanceTests/Common/Program.cs
+++ b/src/PerformanceTests/Common/Program.cs
@@ -93,7 +93,6 @@ namespace Host
             finally
             {
                 Log.Info("Fin!");
-                log4net.LogManager.Shutdown();
             }
             return (int)ReturnCodes.OK;
         }
--- a/src/PerformanceTests/Common/Program.cs
+++ b/src/PerformanceTests/Common/Program.cs
@@ -7,6 +7,7 @@ namespace Host
     using System.Runtime;
     using System.Threading;
     using System.Threading.Tasks;
+    using log4net.Config;
     using Microsoft.Win32;
 #if Version5
     using NServiceBus.Log4Net;
@@ -28,6 +29,8 @@ namespace Host
         {
             Thread.CurrentThread.CurrentCulture = CultureInfo.InvariantCulture;
             Thread.CurrentThread.CurrentUICulture = CultureInfo.InvariantCulture;
+
+            XmlConfigurator.Configure();
 
             return MainAsync().ConfigureAwait(false).GetAwaiter().GetResult();
         }
--- a/src/PerformanceTests/Common/Program.cs
+++ b/src/PerformanceTests/Common/Program.cs
@@ -93,6 +93,7 @@ namespace Host
             finally
             {
                 Log.Info("Fin!");
+                log4net.LogManager.Shutdown();
             }
             return (int)ReturnCodes.OK;
         }
--- a/src/PerformanceTests/Common/Program.cs
+++ b/src/PerformanceTests/Common/Program.cs
@@ -8,7 +8,12 @@ namespace Host
     using System.Threading;
     using System.Threading.Tasks;
     using Microsoft.Win32;
+#if Version5
+    using NServiceBus.Log4Net;
+#else
     using NServiceBus;
+#endif
+
     using NServiceBus.Logging;
     using Tests.Permutations;
     using VisualStudioDebugHelper;
@@ -21,14 +26,16 @@ namespace Host
 
         static int Main()
         {
+            Thread.CurrentThread.CurrentCulture = CultureInfo.InvariantCulture;
+            Thread.CurrentThread.CurrentUICulture = CultureInfo.InvariantCulture;
+
             return MainAsync().ConfigureAwait(false).GetAwaiter().GetResult();
         }
 
         static async Task<int> MainAsync()
         {
             GCSettings.LatencyMode = GCLatencyMode.Batch;
-            LogManager.Use<NLogFactory>();
-            NLog.LogManager.Configuration.DefaultCultureInfo = CultureInfo.InvariantCulture;
+            LogManager.Use<Log4NetFactory>();
 
             Log = LogManager.GetLogger(typeof(Program));
 
@@ -83,7 +90,6 @@ namespace Host
             finally
             {
                 Log.Info("Fin!");
-                NLog.LogManager.Shutdown();
             }
             return (int)ReturnCodes.OK;
         }
--- a/src/PerformanceTests/Common/Scenarios/BaseRunner.v5.cs
+++ b/src/PerformanceTests/Common/Scenarios/BaseRunner.v5.cs
@@ -8,6 +8,7 @@ using NServiceBus;
 using System.Collections.Generic;
 using System.Configuration;
 using System.Threading.Tasks;
+using log4net;
 using NServiceBus.Config;
 using NServiceBus.Config.ConfigurationSource;
 
@@ -85,6 +86,7 @@ partial class BaseRunner
         {
             Log.Fatal("OnCriticalError", exception);
             Session.CloseWithSuppress().ConfigureAwait(false).GetAwaiter().GetResult();
+            LogManager.Shutdown();
         }
         finally
         {
--- a/src/PerformanceTests/Common/Scenarios/BaseRunner.v5.cs
+++ b/src/PerformanceTests/Common/Scenarios/BaseRunner.v5.cs
@@ -83,15 +83,8 @@ partial class BaseRunner
     {
         try
         {
-            try
-            {
-                Log.Fatal("OnCriticalError", exception);
-                Session.CloseWithSuppress().ConfigureAwait(false).GetAwaiter().GetResult();
-            }
-            finally
-            {
-                NLog.LogManager.Shutdown();
-            }
+            Log.Fatal("OnCriticalError", exception);
+            Session.CloseWithSuppress().ConfigureAwait(false).GetAwaiter().GetResult();
         }
         finally
         {
--- a/src/PerformanceTests/Common/Scenarios/BaseRunner.v6.cs
+++ b/src/PerformanceTests/Common/Scenarios/BaseRunner.v6.cs
@@ -87,15 +87,8 @@ partial class BaseRunner
     {
         try
         {
-            try
-            {
-                Log.Fatal("OnCriticalError", context.Exception);
-                await context.Stop().ConfigureAwait(false);
-            }
-            finally
-            {
-                NLog.LogManager.Shutdown();
-            }
+            Log.Fatal("OnCriticalError", context.Exception);
+            await context.Stop().ConfigureAwait(false);
         }
         finally
         {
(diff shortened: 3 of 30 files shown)


src/PerformanceTests/Common/Program.cs
```diff
@@ -83,7 +93,6 @@ static int Main()
             finally
             {
                 Log.Info("Fin!");
-                NLog.LogManager.Shutdown();
```
ramonsmits: I think log4net also has a shutdown method.

src/PerformanceTests/Common/Scenarios/BaseRunner.v5.cs
```diff
@@ -83,15 +83,8 @@ void OnCriticalError(string errorMessage, Exception exception)
     {
         try
         {
-            try
-            {
-                Log.Fatal("OnCriticalError", exception);
-                Session.CloseWithSuppress().ConfigureAwait(false).GetAwaiter().GetResult();
-            }
-            finally
```
ramonsmits: Lgo4net also has a shutdown method.

src/PerformanceTests/Common/Scenarios/BaseRunner.v6.cs
```diff
@@ -87,15 +87,8 @@ IEnumerable<Type> GetTypesToExclude()
...
         try
         {
-            try
-            {
-                Log.Fatal("OnCriticalError", context.Exception);
-                await context.Stop().ConfigureAwait(false);
-            }
-            finally
-            {
-                NLog.LogManager.Shutdown();
```
ramonsmits: log4net also has a shutdown method.

src/PerformanceTests/NServiceBus5/App.config
```diff
@@ -1,26 +1,49 @@
...
+      <mapping>
+        <level value="WARN" />
+        <foreColor value="Purple" />
+      </mapping>
+      <mapping>
+        <level value="FATAL" />
+        <foreColor value="RED" />
+      </mapping>
+      <layout type="log4net.Layout.PatternLayout">
+        <conversionPattern value="%date|%level|%thread|%logger|%message%newline%exception" />
```
ramonsmits: The console logger previously didn't log the exception to keep the information readable.

src/PerformanceTests/NServiceBus5/App.config
```diff
@@ -1,26 +1,49 @@
...
+      </layout>
+    </appender>
+    <appender name="Trace" type="log4net.Appender.TraceAppender" >
+      <layout type="log4net.Layout.PatternLayout">
+        <conversionPattern value="%date|%level|%thread|%logger|%message%newline%exception" />
+      </layout>
+    </appender>
+    <root>
```
ramonsmits: I do not think the current log4net configuration is doing asynchronous logging.

src/PerformanceTests/NServiceBus5/App.config
```diff
@@ -1,26 +1,49 @@
...
 
-  <nlog xmlns="http://www.nlog-project.org/schemas/NLog.xsd" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
-    <targets async="true">
-      <target name="file" xsi:type="File" fileName="trace.log" layout="${longdate:universalTime=true}|${level:uppercase=true}|${threadid:padding=2}|${logger}|${message}${onexception:${newline}${exception:format=tostring}}" />
-      <target name="trace" xsi:type="OutputDebugString" layout="${level:uppercase=true}|${threadid:padding=2}|${logger}|${message}${onexception:${newline}${exception:format=tostring}}" />
-      <target name="console" xsi:type="ColoredConsole" layout="${longdate}|${level:uppercase=true}|${threadid:padding=2}|${logger}|${message}${onexception:${exception:format=message}}" />
-    </targets>
-    <rules>
-      <logger name="NServiceBus.Azure.Transports.WindowsAzureServiceBus.AzureServiceBusQueueCreator" maxlevel="Info" final="true" />
-      <logger name="NServiceBus.Distributor.MSMQ.*" maxlevel="Info" final="true" />
```
ramonsmits: These logger overrides are missing in the log4net configuration.

src/PerformanceTests/NServiceBus6/App.config
```diff
@@ -1,21 +1,46 @@
...
+      <layout type="log4net.Layout.PatternLayout">
+        <conversionPattern value="%date|%level|%thread|%logger|%message%newline%exception" />
+      </layout>
+    </appender>
+    <appender name="Trace" type="log4net.Appender.TraceAppender" >
+      <layout type="log4net.Layout.PatternLayout">
+        <conversionPattern value="%date|%level|%thread|%logger|%message%newline%exception" />
+      </layout>
+    </appender>
+    <root>
```
ramonsmits: Same comments as the v5 one.

src/PerformanceTests/NServiceBus6/App.config
```diff
@@ -1,21 +1,46 @@
...
   </configSections>
 
-  <nlog xmlns="http://www.nlog-project.org/schemas/NLog.xsd" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
-    <targets async="true">
-      <target name="file" xsi:type="File" fileName="trace.log" layout="${longdate:universalTime=true}|${level:uppercase=true}|${threadid:padding=2}|${logger}|${message}${onexception:${newline}${exception:format=tostring}}" />
-      <target name="trace" xsi:type="OutputDebugString" layout="${level:uppercase=true}|${threadid:padding=2}|${logger}|${message}${onexception:${newline}${exception:format=tostring}}" />
-      <target name="console" xsi:type="ColoredConsole" layout="${longdate}|${level:uppercase=true}|${threadid:padding=2}|${logger}|${message}${onexception:${exception:format=message}}" />
-    </targets>
-    <rules>
-      <logger name="NServiceBus.RecoverabilityExecutor" maxlevel="Info" final="true" />
```
ramonsmits: Missing in the log4net version.

src/PerformanceTests/NServiceBus7/App.config
```diff
@@ -2,19 +2,46 @@
...
+        <conversionPattern value="%date|%level|%thread|%logger|%message%newline%exception" />
+      </layout>
+    </appender>
+    <appender name="Trace" type="log4net.Appender.TraceAppender" >
+      <layout type="log4net.Layout.PatternLayout">
+        <conversionPattern value="%date|%level|%thread|%logger|%message%newline%exception" />
+      </layout>
+    </appender>
+    <root>
+      <level value="Debug" />
```
ramonsmits: Same comments as v5 and v6 config.

src/PerformanceTests/Utils/SessionExtensions.cs
```diff
@@ -14,7 +14,7 @@ public static class SessionExtensions
         }
         catch (Exception ex)
         {
-            Log.Warn(ex, "CloseWithSuppress ({0}): {1}", ex.GetType(), ex.Message.Replace(Environment.NewLine,"; "));
+            Log.Warn(ex);
```
ramonsmits: Please add the message as previously build. This helps in filtering the log output.

src/PerformanceTests/Utils/Statistics.cs
```diff
@@ -69,21 +68,18 @@ public static Statistics Instance
     public static IDisposable Initialize(Permutation permutation)
     {
         if (instance != null) throw new InvalidOperationException("Instance already active");
-        return instance = new Statistics(permutation);
+        return instance = new Statistics();
     }
 
-    Statistics(Permutation permutation)
+    Statistics()
```
ramonsmits: Nice!

src/PerformanceTests/Common/Program.cs
timbussmann: There seems to be a LogManager.Shutdown method. The log4net docs is just terrible.

src/PerformanceTests/NServiceBus5/App.config
timbussmann: @ramonsmits it used this before: `${onexception:${exception:format=message}}` which tbh I don't fully understand as this is nlog specific syntax, but it does look like it would output the exception message?

Is this really an issue?

src/PerformanceTests/NServiceBus5/App.config
timbussmann: can you point me towards the log4net docs about asynchronous logging?

src/PerformanceTests/Common/Scenarios/BaseRunner.v5.cs
timbussmann: why would we need to shutdown the logger here? Isn't that done by the finally block in `Program.cs`?

src/PerformanceTests/Utils/SessionExtensions.cs
timbussmann: what does it bring to log the exception + a message which reformats the exception message as well?

src/PerformanceTests/Common/Scenarios/BaseRunner.v5.cs
ramonsmits: The one in Main isn't called when doing `Environment.FailFast`, see it as a process kill.

src/PerformanceTests/NServiceBus5/App.config
ramonsmits: @timbussmann That line only writes the exception message, not the stack trace. This is really convenient when looking at the output during execution. It is not a must but these days with our huge stack traces you do not see anything else on screen.

src/PerformanceTests/NServiceBus5/App.config
ramonsmits: Just searched for it... but log4net doesn't have an async appender OOTB, which sucks. Just leave it as is then. If I run the performance tests and see a noticeable performance dip then we can add an async appender.

src/PerformanceTests/Utils/SessionExtensions.cs
ramonsmits: The location should be present in the log event, "CloseWithSuppress". As said, the formatting is so that the details are on a single line which is useful for filtering. The exception isn't logged on the same line thus when filtering it is not visible. As you can see, the exception message spans multiple lines. Azure is writing info over multiple lines.

If you feel strongly to not do the reformatting, then leave it out but do ass the "CloseWithSuppress" in the log event.

src/PerformanceTests/NServiceBus5/App.config
timbussmann: I don't see a way to emulate this behavior in log4net.

ramonsmits: I would want to see the exception message, but not the stack trace. I just read something about `<IgnoresException value="False" />` but I'm not sure what that does exactly.

src/PerformanceTests/Common/Program.cs
timbussmann: I removed the call to shutdown. It's called by log4net when the appdomain exits. Added it back to the cases where we call Environment.FailFast (although I think we should just call Environment.Fail) instead.

timbussmann:
@ramonsmits addressed your comments.

PR #272, opened 2017-11-06
AmazonSQS for core v5

ramonsmits:


ramonsmits:
The v5 permutations were failing because AmazonSQS wasn't available and it was quick to add. This makes much more tests succeed.

PR #273, opened 2017-11-06
Logging fix and improvements

ramonsmits:


src/PerformanceTests/Common/EnvironmentStats.cs
```diff
@@ -31,5 +32,11 @@ public static void Write()
         log.InfoFormat("EnableDnsRoundRobin: {0}", ServicePointManager.EnableDnsRoundRobin);
         log.InfoFormat("Expect100Continue: {0}", ServicePointManager.Expect100Continue);
         log.InfoFormat("UseNagleAlgorithm: {0}", ServicePointManager.UseNagleAlgorithm);
+
+        log = LogManager.GetLogger("ENVVAR");
+        foreach (var item in Environment.GetEnvironmentVariables().OfType<DictionaryEntry>().OrderBy(x=>x.Key))
+        {
+            log.DebugFormat("{0}={1}", item.Key, item.Value);
```
timbussmann: 👎 this is a security risk

src/PerformanceTests/Common/Program.cs
```diff
@@ -56,12 +56,12 @@ static int Main()
 
                     if (Environment.UserInteractive) Console.Title = PermutationParser.ToFriendlyString(permutation);
 
-                    var runnerTypes = AssemblyScanner.GetAllTypes<BaseRunner>();
+                    var runnerTypes = AssemblyScanner.GetAllTypes<BaseRunner>().ToArray();
 
-                    foreach(var t in runnerTypes) Log.Error(t.FullName);
+                    foreach (var t in runnerTypes) Log.DebugFormat("Found runner: {0}", t.FullName);
```
timbussmann: In core we don't use the shortcut syntax. But this is your call.

The same comment applies for all the if's down below.

src/PerformanceTests/Common/EnvironmentStats.cs
ramonsmits: That is why it is only done with log level DEBUG, as mentioned in the commit message.

timbussmann: is this really necessary?

timbussmann: I'd vote to remove it. +1 on removing the logging of the connection string as well, although that's not the same level of risks, as it only exposes a specific, selected env var and not all other, potential personal env vars on the users machine.

src/PerformanceTests/Common/EnvironmentStats.cs
```diff
@@ -31,5 +32,11 @@ public static void Write()
         log.InfoFormat("EnableDnsRoundRobin: {0}", ServicePointManager.EnableDnsRoundRobin);
         log.InfoFormat("Expect100Continue: {0}", ServicePointManager.Expect100Continue);
         log.InfoFormat("UseNagleAlgorithm: {0}", ServicePointManager.UseNagleAlgorithm);
+
+        log = LogManager.GetLogger("ENVVAR");
```
timbussmann: the GetLogger(string) api is not available when targeting netstandard. Can you switch that, e.g. see https://github.com/Particular/EndToEnd/pull/281

PR #274, opened 2017-11-06
Seeding improvements

ramonsmits:
Seeding improvements:

- Seeding wasn't purely async on v6 and v7. Now it is, except for MSMQ which overrules the batchhelper with an implementation that uses Task.`Run`.
- Draining was broken on v5 resulted in seeding to never happen.
- Sends during seeding previously resulted in the whole test to fail. Sends are expected to sometimes fail, for example due to throttling or reaching a connection limit. Sends during seeding now allow for some retries to happen including exponential backoff. This results in transports like AmazonSQS and SQL Server to be able to complete the seeding operation.

### DIFF
--- a/src/PerformanceTests/Common/Scenarios/BaseRunner.cs
+++ b/src/PerformanceTests/Common/Scenarios/BaseRunner.cs
@@ -127,8 +127,8 @@ public abstract partial class BaseRunner : IContext
                         }
                         catch (OperationCanceledException)
                         {
+                            Interlocked.Decrement(ref count);
                         }
-
                     }).ConfigureAwait(false);
                 Interlocked.Add(ref count, currentBatchSize);
                 var duration = sw.ElapsedMilliseconds;
--- a/src/PerformanceTests/Common/Scenarios/BaseRunner.cs
+++ b/src/PerformanceTests/Common/Scenarios/BaseRunner.cs
@@ -111,7 +111,7 @@ public abstract partial class BaseRunner : IContext
             const int MinimumBatchSeedDuration = 2500;
             var batchSize = 512;
 
-            Log.InfoFormat("BatchHelper type: {0}", BatchHelper.Instance);
+            Log.InfoFormat("BatchHelper type: {0}", BatchHelper.Instance.GetType());
 
             while (!cts.Token.IsCancellationRequested)
             {
--- a/src/PerformanceTests/Common/Scenarios/BaseRunner.cs
+++ b/src/PerformanceTests/Common/Scenarios/BaseRunner.cs
@@ -26,7 +26,6 @@ public abstract partial class BaseRunner : IContext
     protected int MaxConcurrencyLevel { private set; get; }
 
     protected static bool Shutdown { private set; get; }
-    protected readonly BatchHelper.IBatchHelper BatchHelper = global::BatchHelper.Instance;
     protected readonly Statistics Statistics = Statistics.Instance;
 
     public async Task Execute(Permutation permutation, string endpointName)
@@ -112,11 +111,13 @@ public abstract partial class BaseRunner : IContext
             const int MinimumBatchSeedDuration = 2500;
             var batchSize = 512;
 
+            Log.InfoFormat("BatchHelper type: {0}", BatchHelper.Instance);
+
             while (!cts.Token.IsCancellationRequested)
             {
                 var currentBatchSize = batchSize;
                 var sw = Stopwatch.StartNew();
-                await BatchHelper.Batch(currentBatchSize, i => instance.SendMessage(Session)).ConfigureAwait(false);
+                await BatchHelper.Instance.Batch(currentBatchSize, i => instance.SendMessage(Session)).ConfigureAwait(false);
                 Interlocked.Add(ref count, currentBatchSize);
                 var duration = sw.ElapsedMilliseconds;
                 if (duration < MinimumBatchSeedDuration)
--- a/src/PerformanceTests/Common/Scenarios/BaseRunner.cs
+++ b/src/PerformanceTests/Common/Scenarios/BaseRunner.cs
@@ -244,8 +244,8 @@ public abstract partial class BaseRunner : IContext
                 lock (random) next = random.NextDouble();
                 next *= 0.2; // max 20% jitter
                 next += 1D;
-                next *= 100 * Math.Pow(2, ++attempts);
-                var delay = TimeSpan.FromMilliseconds(next);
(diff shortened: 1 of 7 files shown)


ramonsmits:
@timbussmann rebased on master

src/PerformanceTests/Utils/BatchHelper.cs
```diff
@@ -19,6 +19,19 @@ public Task Batch(int count, Func<int, Task> action)
...
     }
+    public class TaskWhenAllTaskRun : IBatchHelper
+    {
+        public Task Batch(int count, Func<int, Task> action)
+        {
+            var sends = new Task[count];
+            for (var i = 0; i < count; i++)
+            {
+                var i2 = i;
+                sends[i] = Task.Run(() => action(i2));
```
timbussmann: this isn't going to perform well for large `count`.

src/PerformanceTests/Common/Scenarios/PerpetualRunner.cs
```diff
@@ -15,6 +15,8 @@ public abstract class PerpetualRunner : BaseRunner
 
         Log.InfoFormat("Trying to seed {0:N0} items within {1:N0}ms...", seedSize, duration);
 
+        Log.InfoFormat("BatchHelper type: {0}", BatchHelper.Instance);
```
timbussmann: did you intend to log the type?

src/PerformanceTests/Common/Scenarios/LoopRunner.cs
```diff
@@ -51,6 +51,8 @@ protected override Task Stop()
 
             const int MinimumBatchSeedDuration = 2500;
 
+            Log.InfoFormat("BatchHelper type: {0}", BatchHelper.Instance);
```
timbussmann: did you intend to log the type?

src/PerformanceTests/Common/Scenarios/BaseRunner.cs
```diff
@@ -214,4 +224,32 @@ protected virtual Task Setup()
...
+                return;
+            }
+            catch (Exception ex)
+            {
+                if (attempts > maxAttempts) throw new InvalidOperationException("Exhausted send retries.", ex);
+                double next;
+                lock (random) next = random.NextDouble();
+                next *= 0.2; // max 20% jitter
+                next += 1D;
+                next *= 100 * Math.Pow(2, ++attempts);
```
timbussmann: that seems to yield quite long delays (of multiple seconds) easily. Is that the intention?

src/PerformanceTests/Common/Scenarios/BaseRunner.cs
```diff
@@ -214,4 +224,32 @@ protected virtual Task Setup()
...
+                await action()
+                    .ConfigureAwait(false);
+                return;
+            }
+            catch (Exception ex)
+            {
+                if (attempts > maxAttempts) throw new InvalidOperationException("Exhausted send retries.", ex);
+                double next;
+                lock (random) next = random.NextDouble();
+                next *= 0.2; // max 20% jitter
```
timbussmann: I don't understand why you want to add that. What's the value of this "jitter"?

ramonsmits: The jitter is to randomize the delays. If you get throttle, imagine 5,000 items get delayed at the same time. If you sleep them all for the same duration you will likely get throttled immediately again. This spreads the retries. Especially over multiple retries.

src/PerformanceTests/Common/Scenarios/BaseRunner.cs
ramonsmits: Yes, that is intentional 100 ms, 200ms, 400ms, 800ms, 1600ms, 3200ms plus a random 0-20% jitter.

src/PerformanceTests/Common/Scenarios/LoopRunner.cs
ramonsmits: Yes, is seeding is slow you can verify in the log which batch sending style is used. I didn't put this in the batch helper itself as that is invoked in a loop which impacts the log size.

src/PerformanceTests/Common/Scenarios/PerpetualRunner.cs
ramonsmits: Yes, shall I add a code comment?

src/PerformanceTests/Common/Scenarios/BaseRunner.cs
ramonsmits: Shall I add a code comment?

src/PerformanceTests/Utils/BatchHelper.cs
ramonsmits: Could you define large? Tested this with size of 150,000 and worked flawlessly, see https://github.com/Particular/NServiceBus.Transport.Msmq/pull/31

src/PerformanceTests/Common/Scenarios/LoopRunner.cs
timbussmann: Personally, I'd prefer if you make this explicit by passing the type instead of an object as the parameter to not rely on the actual `ToString` implementation

src/PerformanceTests/Common/Scenarios/BaseRunner.cs
timbussmann: >  100 ms, 200ms, 400ms, 800ms, 1600ms, 3200ms plus a random 0-20% jitter.

then shouldn't this be `attempts++` instead of `++attempts`?

src/PerformanceTests/Utils/BatchHelper.cs
timbussmann: I think the overhead generated by that many task.run's won't be that effective as more refined solutions (e.g. https://docs.particular.net/nservicebus/handlers/async-handlers#concurrency-large-amount-of-concurrent-message-operations) . But since this is non-performance relevant code, I guess that's good enough.

src/PerformanceTests/Common/Scenarios/BaseRunner.cs
ramonsmits: Let's see, attempts is 0, gets increased, Math.Path(2,1) results in 2, so yes, good catch @timbussmann

src/PerformanceTests/Common/Scenarios/LoopRunner.cs
ramonsmits: Good suggestion, that better indicates intent.

PR #279, opened 2017-11-09
Make performance counters conditional

timbussmann:
@Particular/endtoend-maintainers performance counters are currently not supported on .net core (and if, only on windows) so I think we need to remove them from the performance tests altogether. It doesn't seem like these performance counters are actually read somewhere automatically, or did I miss something?

Maybe there are ways to configure alternative sinks we can log metrics to in the future, but that seems to be out of scope for now?


ramonsmits:
Lets keep this in with a conditional compile.

danielmarbach:
@ramonsmits why?

ramonsmits:
@danielmarbach Because they are used during testing to get real-time feedback.

timbussmann:
part of https://github.com/Particular/EndToEnd/pull/283 now

PR #281, opened 2017-11-09
Target netstandard in Utils

timbussmann:
GetLogger(string) is not supported by log4net on the netstandard API. Another approach would be to use `GetLogger(Assembly.GetCallingAssembly(), name)`, but I think the typeof approach should work fine as well.


timbussmann:
this one is ready for review as well.

PR #282, opened 2017-11-10
Remove x86 permutation option

timbussmann:
* Gets rid of all the magic to create x86 targets of the exes. So far it seems that none of the tests uses this permutation option anyway.

I kept the permutation variable around (but restricted to x64) to change it to the actual .NET platform targeted (.net core/.net framework).


ramonsmits:
PlatformFixture does all the permutations of 32/64 bit and different garbage collector configurations.

ramonsmits:
While I agree that this is not needed to compare v6 to v7 the performance test are not only about comparing 5, 6 and 7. Besides, .net core has both a 32 and 64 distributions. I'm fine for now that we only do x64 .net core but not sure why this 'magic' needs to be removed.

timbussmann:
In what scenarios would customers really use 32 bit systems? I'd say there is only very little value in testing 32 bit versions explicitly against 64 bit version.

ramonsmits:
@timbussmann Running as 32 is all about memory usage. Most applications do not require 64 bit addressing meaning all pointers are half of their size. If you do not need the memory space there is no reason to run as 64 bit. It is not about running on 32 bit systems, its all about executing it as 32 bit to consume less system resources even on 64 bit systems.

https://stackoverflow.com/questions/6512733/x64-vs-x86-performance-considerations-net

Furthermore, the results show that running x86 with client garbage collector results in 3x less memory usage then x64 server GC.

I am not in favor of removing this test as it indicates these big differences.

timbussmann:
@ramonsmits putting the discussion about the value of this aside, I'm not sure it's something we can support in core. The current approach using CorFlags.exe doesn't work with .NET Core. Do you or maybe @bording have an idea how this could be solved?

ramonsmits:
@timbussmann With .net core you have a 32 installable and a 64 bit installable to run your .net stuff.

timbussmann:
@ramonsmits and can you have them installed side by side?

bording:
On Windows, you can install both the 32bit and 64bit SDKs, but you're only going to have one of them on your path, so you'd have to explicitly call the correct dotnet.exe to get the bitness you want.

Instead of relying on needing both SDKs, you could consider publishing the app as a self-contained deployment, and then you get an exe with the correct bitness based on the RID you specify when you publish. That would rather greatly increase the size of the artifacts though, since it includes the entire set of .NET Core assemblies.

Something else to consider is that for macOS and Linux, there's only one choice, and that is 64bit. There aren't 32bit version of .NET Core available for those platforms.

Linux distros only support a single bitness and can't run both like Windows can. macOS will be phasing out an sort of 32bit support very soon as well.

32bit is becoming increasingly irrelevant, so I'm not sure it's worth the extra complexity to support both here.

Something else to keep in mind is that I know all of the build agents only have the 64bit SDK installed.

timbussmann:
thanks for this awesome summary @bording ! Based on the complexity added to run 32 vs. 64 bit, I still vote to continue with this PR and drop this permutation option. While there might be performance differences, I think it's not worth to maintain the infrastructure to automatically test this, especially as this is usually not a conscious user decision to run on 32 or 64 bit mode. If at some point this is something we want to keep track off, I think this should be handled via pdev instead.

danielmarbach:
I agree with @bording and @timbussmann 

Just to add another data point. Service Fabric is 64 bit only.

ramonsmits:
@bording Thanks for that detailed comment. Lets get this in then. @timbussmann I request only a small change and that is `Platform.x64` to be renamed to `Platform.net452`, if you also want to still include x64 that is fine by me. I leave that up to you. Does that make sense?

timbussmann:
@ramonsmits will do that. Will rebase this PR on top of https://github.com/Particular/EndToEnd/pull/283 and ping again when ready

timbussmann:
@ramonsmits @danielmarbach @bording rebased this PR and applies Ramon's suggestions. This one is ready for review.

PR #286, opened 2017-11-13
Azure Storage Queues, not specifying transaction mode for send-only endpoint

ramonsmits:


src/PerformanceTests/Transport.V7.AzureStorageQueues.V8/AzureStorageQueuesProfile.cs
```diff
@@ -24,7 +24,7 @@ public void Configure(EndpointConfiguration endpointConfiguration)
             && Permutation.TransactionMode != TransactionMode.Receive
             ) throw new NotSupportedException("TransactionMode: " + Permutation.TransactionMode);
 
-        if (Permutation.TransactionMode != TransactionMode.Default) transport.Transactions(Permutation.GetTransactionMode());
+        if (Permutation.TransactionMode != TransactionMode.Default && !Context.IsSendOnly) transport.Transactions(Permutation.GetTransactionMode());
```
danielmarbach: Shouldn't this be consistently applied to all profiles targetting v7 of NSB?

ramonsmits: @danielmarbach ASQ is the only transport that doesn't allow this but I get your thought. Maybe after this patch is now does work again?
 I also don't know if that must be done. For example, what if the transport supports MSDTC and the customer uses transactionscope wouldn't the selected transaction mode make sense?

ramonsmits: Yes, not needed at all, closing this.

PR #294, opened 2017-11-14
Fix file copy sources

timbussmann:
as we're already multitargeting, the current file copy mechanism copies files from both targets into the test execution folder, causing weird assembly loading issues.

ping @ramonsmits

### DIFF
--- a/src/PerformanceTests/Tests/Tools/PermutationDirectoryResolver.cs
+++ b/src/PerformanceTests/Tests/Tools/PermutationDirectoryResolver.cs
@@ -28,19 +28,32 @@ namespace Tests.Tools
             if (permutation.Version == NServiceBusVersion.V7)
             {
                 // v7 targeting projects are using the new csproj format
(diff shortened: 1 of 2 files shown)


src/PerformanceTests/Tests/Tools/PermutationDirectoryResolver.cs
```diff
@@ -49,10 +62,25 @@ string GetRootDirectory(NServiceBusVersion version)
...
+        {
+            switch (permutation.Version)
+            {
+                case NServiceBusVersion.V5:
+                case NServiceBusVersion.V6:
+                    return GetHostName(permutation);
+                case NServiceBusVersion.V7:
+                    return Path.Combine(GetHostName(permutation), GetTargetFrameworkName(permutation));
+                default:
+                    throw new NotSupportedException(permutation.Version.ToString("G"));
```
ramonsmits: Just learned that enums have ToString overload :-)

PR #295, opened 2017-11-14
Connectionstrings outside repo

ramonsmits:
Host for v7 contained a double entry which is removed. The host projects now link to a connection string file outside of the repo. This way we can perform a future build on linux too but custom connection string values wil not indicate a change in the repo.


ramonsmits:
@timbussmann Please ack this is you agree with the solution that we discussed.

PR #297, opened 2017-11-16
Support .NET Core as platform

timbussmann:
This is in a super early WIP state.

### DIFF
--- a/src/PerformanceTests/Common/CheckGCServer.cs
+++ b/src/PerformanceTests/Common/CheckGCServer.cs
@@ -14,7 +14,7 @@ class CheckGCServer : IProfile, INeedPermutation
 
     public void Configure(Configuration cfg)
     {
-        if (Permutation.GarbageCollector == GarbageCollector.Client && GCSettings.IsServerGC) throw new InvalidOperationException("GarbageCollector is set to Server but must be Client.");
-        if (Permutation.GarbageCollector == GarbageCollector.Server && !GCSettings.IsServerGC) throw new InvalidOperationException("GarbageCollector is set to Client but must be Server.");
+        //if (Permutation.GarbageCollector == GarbageCollector.Client && GCSettings.IsServerGC) throw new InvalidOperationException("GarbageCollector is set to Server but must be Client.");
+        //if (Permutation.GarbageCollector == GarbageCollector.Server && !GCSettings.IsServerGC) throw new InvalidOperationException("GarbageCollector is set to Client but must be Server.");
     }
 }
--- a/src/PerformanceTests/Common/CheckGCServer.cs
+++ b/src/PerformanceTests/Common/CheckGCServer.cs
@@ -14,7 +14,12 @@ class CheckGCServer : IProfile, INeedPermutation
 
     public void Configure(Configuration cfg)
     {
-        //if (Permutation.GarbageCollector == GarbageCollector.Client && GCSettings.IsServerGC) throw new InvalidOperationException("GarbageCollector is set to Server but must be Client.");
-        //if (Permutation.GarbageCollector == GarbageCollector.Server && !GCSettings.IsServerGC) throw new InvalidOperationException("GarbageCollector is set to Client but must be Server.");
+        if (Permutation.Platform == Platform.NetFramework)
+        {
+            if (Permutation.GarbageCollector == GarbageCollector.Client && GCSettings.IsServerGC) throw new InvalidOperationException("GarbageCollector is set to Server but must be Client.");
+            if (Permutation.GarbageCollector == GarbageCollector.Server && !GCSettings.IsServerGC) throw new InvalidOperationException("GarbageCollector is set to Client but must be Server.");
+        }
+        
+        // TODO: GC cannot only be configured via csproj in .net core. Should we throw an exception if the GC mode doesn't match?
     }
 }
--- a/src/PerformanceTests/Common/Program.cs
+++ b/src/PerformanceTests/Common/Program.cs
@@ -30,27 +30,17 @@ namespace Host
 
         static int Main()
         {
-            Console.WriteLine("starting");
-            Console.WriteLine("BaseDir:" + AppDomain.CurrentDomain.BaseDirectory);
-            Console.WriteLine("CurrentDir:" + Environment.CurrentDirectory);
-            Console.WriteLine("rel search path:" + AppDomain.CurrentDomain.RelativeSearchPath);
-            Thread.CurrentThread.CurrentCulture = CultureInfo.InvariantCulture;
-            Thread.CurrentThread.CurrentUICulture = CultureInfo.InvariantCulture;
+            var entryAssembly = Assembly.GetEntryAssembly();
 
-            Console.WriteLine("executing assembly: " + Assembly.GetExecutingAssembly());
-//            Console.WriteLine(ConfigurationManager.GetSection("log4net"));
-
-
-
-
-            XmlConfigurator.Configure(log4net.LogManager.GetRepository(Assembly.GetExecutingAssembly()));
-
+            var logRepository = log4net.LogManager.GetRepository(entryAssembly);
+            string configFileName = Path.GetFileName(new Uri(entryAssembly.CodeBase).LocalPath);
+            XmlConfigurator.Configure(logRepository, new FileInfo($"{configFileName}.config"));
+            
             return MainAsync().ConfigureAwait(false).GetAwaiter().GetResult();
         }
 
         static async Task<int> MainAsync()
         {
-            Console.WriteLine("main async");
             GCSettings.LatencyMode = GCLatencyMode.Batch; // https://docs.microsoft.com/en-us/dotnet/standard/garbage-collection/latency
             LogManager.Use<Log4NetFactory>();
 
--- a/src/PerformanceTests/Common/Program.cs
+++ b/src/PerformanceTests/Common/Program.cs
@@ -30,6 +30,10 @@ namespace Host
 
         static int Main()
         {
+            ServicePointManager.Expect100Continue = false;
+            ServicePointManager.UseNagleAlgorithm = false;
+            ServicePointManager.DefaultConnectionLimit = 150;
+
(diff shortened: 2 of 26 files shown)


src/PerformanceTests/Common/Statistics/SimpleStatisticsFeature.V6.cs
```diff
@@ -22,6 +22,8 @@ public SimpleStatisticsFeature()
 
         protected override void Setup(FeatureConfigurationContext context)
         {
+            Console.WriteLine("setup SimpleStatisticsFeature");
```
timbussmann: I see this setup being called twice in the `GatedSendLocalRunner` test, is that expected @ramonsmits ?

src/PerformanceTests/NServiceBus7/App.config
```diff
@@ -106,24 +89,4 @@
...
-  <system.net>
-    <settings>
-      <servicePointManager expect100Continue="false" useNagleAlgorithm="false" />
-    </settings>
-    <connectionManagement>
-      <add address="*" maxconnection="150" />
-    </connectionManagement>
-  </system.net>
-  <system.web>
-    <membership defaultProvider="ClientAuthenticationMembershipProvider">
```
timbussmann: not sure what these would be required for?

src/PerformanceTests/Common/Program.cs
```diff
@@ -28,11 +30,18 @@ class Program
...
+
+            var entryAssembly = Assembly.GetEntryAssembly();
             Thread.CurrentThread.CurrentCulture = CultureInfo.InvariantCulture;
             Thread.CurrentThread.CurrentUICulture = CultureInfo.InvariantCulture;
 
-            XmlConfigurator.Configure(log4net.LogManager.GetRepository(Assembly.GetExecutingAssembly()));
-
+            var logRepository = log4net.LogManager.GetRepository(entryAssembly);
+            var configFileName = Path.GetFileName(new Uri(entryAssembly.CodeBase).LocalPath);
+            XmlConfigurator.Configure(logRepository, new FileInfo($"{configFileName}.config"));
```
timbussmann: @danielmarbach do we even need the repository argument?

danielmarbach: Good question, no idea... It works (TM) ;)

src/PerformanceTests/Common/CheckGCServer.cs
```diff
@@ -14,7 +14,12 @@ class CheckGCServer : IProfile, INeedPermutation
...
     {
-        if (Permutation.GarbageCollector == GarbageCollector.Client && GCSettings.IsServerGC) throw new InvalidOperationException("GarbageCollector is set to Server but must be Client.");
-        if (Permutation.GarbageCollector == GarbageCollector.Server && !GCSettings.IsServerGC) throw new InvalidOperationException("GarbageCollector is set to Client but must be Server.");
+        if (Permutation.Platform == Platform.NetFramework)
+        {
+            if (Permutation.GarbageCollector == GarbageCollector.Client && GCSettings.IsServerGC) throw new InvalidOperationException("GarbageCollector is set to Server but must be Client.");
+            if (Permutation.GarbageCollector == GarbageCollector.Server && !GCSettings.IsServerGC) throw new InvalidOperationException("GarbageCollector is set to Client but must be Server.");
+        }
+        
+        // TODO: GC cannot only be configured via csproj in .net core. Should we throw an exception if the GC mode doesn't match?
```
danielmarbach: @ramonsmits are you ok if we let this in here for now?

src/PerformanceTests/Common/Statistics/SimpleStatisticsFeature.V6.cs
danielmarbach: @timbussmann let's remove this?

src/PerformanceTests/Common/Statistics/SimpleStatisticsFeature.V6.cs
```diff
@@ -22,6 +22,8 @@ public SimpleStatisticsFeature()
...
+            Console.WriteLine(this.GetType().AssemblyQualifiedName);
```
danielmarbach: @timbussmann let's remove this?

src/PerformanceTests/Utils/AssemblyScanner.cs
```diff
@@ -14,7 +14,15 @@ public static IEnumerable<Assembly> GetAssemblies()
 
         foreach (var path in Directory.EnumerateFiles(Environment.CurrentDirectory, "*.dll"))
         {
-            Assembly.LoadFile(path);
+            var assemblyName = Path.GetFileNameWithoutExtension(path);
+            try
+            {
+                Assembly.LoadFrom(path);
```
danielmarbach: Should we remove the try catch and just let it bomb here?

src/PerformanceTests/Common/CheckGCServer.cs
ramonsmits: @danielmarbach With .net framework GC was also only configurable via the config. The test runner patches the XML for it.

danielmarbach: Ok then I can fix it

ramonsmits: I would write a log entry with level WARN to identify that this setting is ignored with .net core if we cannot patch the config as is done for .net framework.

src/PerformanceTests/Common/Statistics/SimpleStatisticsFeature.V6.cs
ramonsmits: @timbussmann The endpoint is created multiple times so that is probably the reason you see it.

ramonsmits: Not sure why this was added? I think these write lines can be removed.

src/PerformanceTests/Common/CheckGCServer.cs
danielmarbach: Scratch that comment. Is patches the App.Config but we would need to patch the csproj which does not sound feasible?

src/PerformanceTests/NServiceBus7/App.config
```diff
@@ -6,19 +6,7 @@
...
-      </mapping>
-      <mapping>
-        <level value="WARN" />
-        <foreColor value="Purple" />
-      </mapping>
-      <mapping>
-        <level value="FATAL" />
-        <foreColor value="RED" />
-      </mapping>
+    <appender name="Console" type="log4net.Appender.ConsoleAppender" >
```
ramonsmits: No colored console appender in .net core?

src/PerformanceTests/NServiceBus7/App.config
ramonsmits: I'm not sure what the `system.web` stuff is for, I would remove it and I'll see if one of transports require it but I think not. Its different for the `system.net` section, that must remain.

src/PerformanceTests/NServiceBus7/App.config
```diff
@@ -106,24 +89,4 @@
     </assemblyBinding>
     <gcServer enabled="false" />
   </runtime>
-  <system.net>
```
ramonsmits: Do not remove `system.net`.

ramonsmits: O wait, I see, you've swapped it for code API.

src/PerformanceTests/NServiceBus7/App.config
danielmarbach: Nope not supported. http://logging.apache.org/log4net/release/framework-support.html#appenders

src/PerformanceTests/Common/CheckGCServer.cs
timbussmann: yeah, the configuration happens in the csproj file, that's not patchable. It might be possible to patch the the `runtimeconfig.json` file (but not with the current XML patching obviously).

src/PerformanceTests/Utils/AssemblyScanner.cs
timbussmann: I had Load fail when trying to load a runtime specific dll for the sql client package. As this assembly isn't in the test directory anymore, we should be good to remove the try catch

src/PerformanceTests/Common/Statistics/SimpleStatisticsFeature.V6.cs
timbussmann: yeah, some debugging leftovers

src/PerformanceTests/Common/Statistics/SimpleStatisticsFeature.V6.cs
timbussmann: some debugging leftovers

src/PerformanceTests/NServiceBus7/App.config
timbussmann: dotnet core's configurationmanager package isn't able to load system.net sections and will throw.

src/PerformanceTests/NServiceBus7/App.config
ramonsmits: well, then we must live without any color :-)

src/PerformanceTests/Common/CheckGCServer.cs
ramonsmits: I only found this snippet

ramonsmits: Also reading something about setting `complus_gcserver` env var, probably that could be done too for example in the batch file and also in the runner but patching that json file with just a string.Replace seems pretty trivial.

danielmarbach: >All settings in this group, except for the "System.GC.Server" property, are placed into a file called runtimeconfig.template.json

https://docs.microsoft.com/en-us/dotnet/core/tools/project-json-to-csproj

😭

PR #298, opened 2017-11-17
Tests filtered out on .net core for older NServiceBus versions.

ramonsmits:


src/PerformanceTests/Utils/Permutations/PermutationGenerator.cs
```diff
@@ -54,6 +55,9 @@ public static IEnumerable<Permutation> Generate(Permutations permutations, Func<
                      + (permutations.ScaleOuts.Length > 1 ? ScaleOut + Separator : string.Empty)
                 };
 
+            // 7+ only supports .net core
+            items = items.Where(x => x.Platform == Platform.NetCore && x.Version == NServiceBusVersion.V7 || x.Platform != Platform.NetCore);
```
danielmarbach: Since the underlying value of the version enum is an int could we use `>=`?

ramonsmits: @danielmarbach good idea, I'll assign explicit values if not already set.

PR #300, opened 2017-11-21
[SPIKE] Test Harness .NET Core

danielmarbach:
Connects to https://github.com/Particular/DotNetCoreLaunch/issues/71


src/PerformanceTests/Tests/Categories/Base.cs
```diff
@@ -95,20 +113,23 @@ static void LaunchAndWait(Permutation permutation, TestDescriptor testDescriptor
...
+#if NET452
                 var processId = DebugAttacher.GetCurrentVisualStudioProcessId();
-                var processIdArgument = processId >= 0 ? $" --processId={processId}" : string.Empty;
-
+                processIdArgument = processId >= 0 ? $" --processId={processId}" : string.Empty;
+#endif
                 pi = new ProcessStartInfo(testDescriptor.ProjectAssemblyPath, permutationArgs + sessionIdArgument + processIdArgument)
                 {
-                    UseShellExecute = false,
+                    UseShellExecute = true,
```
ramonsmits: This is probably going to cause issues as now the test will be launched in a new console.

src/PerformanceTests/Tests/Tools/TestEnvironment.cs
```diff
@@ -153,7 +164,7 @@ static void Clone(string src, string dst)
                 if (equalTimestamps) return;
                 File.Delete(dst);
             }
-            if (!SymbolicLink.Create(src, dst)) File.Copy(src, dst);
+            File.Copy(src, dst);
```
ramonsmits: Not using symlinks will require large amounts of storage if you are going to run all tests!!

src/PerformanceTests/Utils/Utils.csproj
```diff
@@ -15,6 +15,7 @@
   <ItemGroup>
     <PackageReference Include="CommandLineParser" Version="2.1.1-beta" />
     <PackageReference Include="log4net" Version="2.0.8" />
+    <PackageReference Include="Microsoft.DotNet.Analyzers.Compatibility" Version="0.1.2-alpha" />
```
ramonsmits: Nice!

src/PerformanceTests/Tests/Categories/Base.cs
```diff
@@ -10,13 +10,31 @@ namespace Categories
...
-        static readonly bool InvokeEnabled = bool.Parse(ConfigurationManager.AppSettings["InvokeEnabled"]);
-        static readonly TimeSpan MaxDuration = TimeSpan.Parse(ConfigurationManager.AppSettings["MaxDuration"]);
+        static readonly bool InvokeEnabled;
+        static readonly TimeSpan MaxDuration;
+
+        static Base()
+        {
+            try
+            {
+//                InvokeEnabled = bool.Parse(ConfigurationManager.AppSettings["InvokeEnabled"]);
```
timbussmann: tbh I don't really know why this isn't working. I expected this to work as we're using the config manager package but it doesn't even work when running the tests on .net core under windows. Any ideas?

src/PerformanceTests/Tests/Categories/Base.cs
timbussmann: I had to enable this when running the test (not the endpoint) on .net core, as there was no console showing up at all. This might be something we have to #ifdef as well.

src/PerformanceTests/Tests/Tools/TestEnvironment.cs
timbussmann: @ramonsmits but the number of tests & permutations which are capable of running on linux/.net core is fairly small.

ramonsmits: @timbussmann If you run all tests on V5,V6,V7, and V7 dotnet core, that is alot of data. I agree that is you only do a subset then it isn't really an issue.

ramonsmits: @danielmarbach Is that ` ReparsePointUtility` available on net452 too?

ramonsmits: I would be in favor of just cloning that method, although I don't know how it would perform but probably fast enough compared to an API call.

ramonsmits: Also, this is part of the test runner, do we want that to be able to run on dotnet core?

timbussmann: > Also, this is part of the test runner, do we want that to be able to run on dotnet core?

if we want to run the tests on linux we kinda have to unless you want to involve a lot of manual steps.

ramonsmits: A lot of manual steps? Probably just the creation of a script that does a recursive search on NServiceBus?.dll/exe and then execute it or maybe I'm simpllfying it too much but it we can use the test runner that actually can generate the permutations that would surely be really nice.

timbussmann: without dotnet core support you would have to generate the permutation output on a windows machine before copying files over to platform you want to run as you already said, that makes for a very annoying testing process. 
Additionally you would have to make sure that all the dependencies (nuget packages) are available, which might be tricky, given that you cannot use dotnet restore (as you don't have a csproj available).

src/PerformanceTests/Common/EnvironmentStats.cs
```diff
@@ -34,9 +33,9 @@ public static void Write()
         log.InfoFormat("UseNagleAlgorithm: {0}", ServicePointManager.UseNagleAlgorithm);
 
         log = LogManager.GetLogger("ENVVAR");
-        foreach (var item in Environment.GetEnvironmentVariables().OfType<DictionaryEntry>().OrderBy(x=>x.Key))
+        foreach (var key in Environment.GetEnvironmentVariables().Keys.OfType<string>().OrderBy(x=> x))
         {
-            log.DebugFormat("- {0}", item.Key);
+            log.InfoFormat("{0}: {1}", key, Environment.GetEnvironmentVariable(key));
```
timbussmann: currently, it doesn't log the value for security reasons, so just using the string should be good enough it seems?

src/PerformanceTests/Tests/Categories/Base.cs
timbussmann: I found this issue and it seems to be the same issue: https://github.com/dotnet/corefx/issues/22101 as the entry assembly here is testhost.dll as well.

src/PerformanceTests/Tests/Categories/Base.cs
timbussmann: testing this a bit more explicitly, it doesn't seem to cause any issues or spawn new consoles when running this with UseShellExecute = true.

danielmarbach:
Given that #306 and #307 are merged I'm closing this

PR #302, opened 2017-11-22
Logging progress of seeding of invidual batches

ramonsmits:
Based on feedback by @timbussmann that asked a question about the following log entries and thought that after the batch size increment only a single batch of messages was seeded.
```
2017-11-22 09:04:44,212|INFO|11|BaseRunner|Increasing seed batch size to   2,048 as sending took   1,884ms which is less then 2,500ms
2017-11-22 09:05:08,066|INFO|18|BaseRunner|Done seeding, seeded 13,824 messages, 516.3 msg/s
```

### DIFF
--- a/src/PerformanceTests/Common/Scenarios/BaseRunner.cs
+++ b/src/PerformanceTests/Common/Scenarios/BaseRunner.cs
@@ -132,7 +132,7 @@ public abstract partial class BaseRunner : IContext
                     }).ConfigureAwait(false);
                 Interlocked.Add(ref count, currentBatchSize);
                 var duration = sw.ElapsedMilliseconds;
-                Log.InfoFormat("Batch with size {0,7:N0} duration {1,7:N0}ms", batchSize, duration);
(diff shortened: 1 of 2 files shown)


src/PerformanceTests/Common/Scenarios/BaseRunner.cs
```diff
@@ -132,6 +132,7 @@ await RetryWithBackoff(() => instance.SendMessage(Session), cts.Token, i.ToStrin
                     }).ConfigureAwait(false);
                 Interlocked.Add(ref count, currentBatchSize);
                 var duration = sw.ElapsedMilliseconds;
+                Log.InfoFormat("Batch with size {0,7:N0} duration {1,7:N0}ms", batchSize, duration);
```
timbussmann: maybe "Batch completed with [...]" ?

ramonsmits:
@timbussmann done, merged it with that comment.

PR #303, opened 2017-11-24
DefaultConnectionLimit from ServicePointManager is not taking into account with .NET Core

danielmarbach:


danielmarbach:
I tried to implement it by querying the ServicePointManager first but that didn't work so I introduced the constant

ramonsmits:
@danielmarbach Do you mean that if you first assign 150 to ServicePointManager.DefaultConnectionLimit and then query it that it does not return 150?

src/PerformanceTests/Common/Program.cs
```diff
@@ -30,7 +30,7 @@ static int Main()
         {
             ServicePointManager.Expect100Continue = false;
             ServicePointManager.UseNagleAlgorithm = false;
-            ServicePointManager.DefaultConnectionLimit = 150;
+            ServicePointManager.DefaultConnectionLimit = Settings.DefaultConnectionLimit;
```
ramonsmits: In this case a code comment on why this is done would help to prevent someone refactoring this in the future.

danielmarbach:
@ramonsmits I meant

ServicePointManager.DefaultConnectionLimit = 150;
#if !NET452
                MaxConnectionsPerServer = ServicePointManager.DefaultConnectionLimit;
#endif

did not work

src/PerformanceTests/Common/Program.cs
danielmarbach: Feel free to add a comment if you think it is important.

PR #304, opened 2017-11-27
NSB 7 Host Unix compatibility

timbussmann:
relates to https://github.com/Particular/DotNetCoreLaunch/issues/71
addresses the non unix compatible APIs used in the Host of the performance tests as shown in https://github.com/Particular/EndToEnd/pull/300

Kept all the logic in place in case the .net core application runs on windows.

### DIFF
--- a/src/PerformanceTests/Common/Statistics/SimpleStatisticsFeature.V6.cs
+++ b/src/PerformanceTests/Common/Statistics/SimpleStatisticsFeature.V6.cs
@@ -57,7 +57,7 @@ namespace NServiceBus.Performance
             private static readonly long TimeUnit = Stopwatch.Frequency;
 
             private readonly string _titleFormat;
-            private readonly bool _outputTitle;
+            private readonly bool _outputTitle = Environment.OSVersion.Platform == PlatformID.Win32NT;
             private readonly bool _outputLog;
             private readonly int _updateInMilliSeconds = 2000;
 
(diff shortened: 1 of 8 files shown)


src/PerformanceTests/Common/Statistics/SimpleStatisticsFeature.V6.cs
```diff
@@ -57,7 +57,7 @@ public class Collector : StatisticsBehavior.Implementation
             private static readonly long TimeUnit = Stopwatch.Frequency;
 
             private readonly string _titleFormat;
-            private readonly bool _outputTitle;
+            private readonly bool _outputTitle = Environment.OSVersion.Platform == PlatformID.Win32NT;
```
ramonsmits: I think this also needs `&& Environment.UserInteractive` as you did elsewhere. It probably is a bug that it previously was always set to true.

danielmarbach: @ramonsmits do we ever run those tests in non-interactive mode? I'm asking for me learning

timbussmann: what would happen if you would set the title when not in user interactive mode?

ramonsmits: @danielmarbach I don't think so, I don't see the need to run this as a windows service. This code originally is from another project.

@timbussmann Trying to set the console results in an exception as there is no console when you run as a windows service.

timbussmann: @ramonsmits given that explanation, I think the current code should be fine then?

PR #305, opened 2017-11-28
Utils project Unix compatibility

timbussmann:
relates to relates to https://github.com/Particular/DotNetCoreLaunch/issues/71
Contains the utils specific changes shown in https://github.com/Particular/EndToEnd/pull/300

WIP as it's currently built on top of https://github.com/Particular/EndToEnd/pull/304


src/PerformanceTests/Tests/Tools/TestEnvironment.cs
```diff
@@ -153,7 +153,8 @@ static void Clone(string src, string dst)
                 if (equalTimestamps) return;
                 File.Delete(dst);
             }
-            if (!SymbolicLink.Create(src, dst)) File.Copy(src, dst);
```
timbussmann: SymbolicLink.Create never returned true on my machine so I removed it. Is there something required for symlinks to work? @ramonsmits

danielmarbach: Should we set the override flag to true https://msdn.microsoft.com/en-us/library/9706cfs5(v=vs.110).aspx?

ramonsmits: @timbussmann You have to run with elevated permissions

- https://github.com/Particular/EndToEnd/blob/master/src/PerformanceTests/readme.md#run-visual-studio-as-administrator-to-save-disk-space

ramonsmits: @danielmarbach The code already does a `File.Exists(dst)` so that boolean isn't required.

src/PerformanceTests/Tests/Tools/TestEnvironment.cs
```diff
@@ -153,7 +153,8 @@ static void Clone(string src, string dst)
...
+
+            File.Copy(src, dst);
```
ramonsmits: @danielmarbach You probably meant this `File.Copy`, yes, here it probably is a good idea to allow overwrites.

src/PerformanceTests/Utils/Powerplan.cs
```diff
@@ -1,9 +1,40 @@
...
 using System.Runtime.InteropServices;
+using System.Threading;
+using log4net;
 
 public class Powerplan
 {
-    public static Guid GetActive()
+    static ILog Log = LogManager.GetLogger(typeof(Powerplan));
+
+    public static void CheckPowerPlan()
```
ramonsmits: I'm fine with moving this, although this method doesn't really belong to the Powerplan class.

timbussmann: I partially agree :D In general, the check should indeed not belong to the powerplan class, but as long as the check contains code like this:
> var highperformance = new Guid("8c5e7fda-e8bf-4a96-9a85-a6e23a8c635c");
it's leaking too many internals and I don't think this should be outside the powerplan class. If the Powerplan class would have a property like `Powerplan.IsSetToHighPerformance` which can easily be checked without requiring knowledge of magic guids, I'd totally agree with you. For now, I'd suggest to keep as is and see whether we can get a round of refactorings in before closing the issue.

src/PerformanceTests/Tests/Tools/TestEnvironment.cs
timbussmann: ah I missed the part about required elevated permissions, given that condition, I am even more inclined to remove this code

PR #306, opened 2017-11-29
Multitarget tests

timbussmann:


src/PerformanceTests/Tests/Categories/Base.cs
```diff
@@ -10,13 +10,22 @@ namespace Categories
...
-        static readonly bool InvokeEnabled = bool.Parse(ConfigurationManager.AppSettings["InvokeEnabled"]);
-        static readonly TimeSpan MaxDuration = TimeSpan.Parse(ConfigurationManager.AppSettings["MaxDuration"]);
+        static readonly bool InvokeEnabled;
+        static readonly TimeSpan MaxDuration;
+
+        static Base()
+        {
+            // we need to get the correct config file, as .net core will try to load nunits test harness configuration file
+            // see https://github.com/dotnet/corefx/issues/22101
+            var assemblyLocation = Assembly.GetExecutingAssembly().Location;
```
danielmarbach: I was wondering if it wouldn't be simpler to replace the config file with a code first approach. @ramonsmits how often are these values changed?

src/PerformanceTests/Tests/Categories/Base.cs
```diff
@@ -92,23 +101,27 @@ static void LaunchAndWait(Permutation permutation, TestDescriptor testDescriptor
...
+                // ReSharper disable once RedundantAssignment
+                var processIdArgument = string.Empty;
+#if NET452
+                var processId = VisualStudioDebugHelper.DebugAttacher.GetCurrentVisualStudioProcessId();
+                processIdArgument = processId >= 0 ? $" --processId={processId}" : string.Empty;
+#endif
                 pi = new ProcessStartInfo(testDescriptor.ProjectAssemblyPath, permutationArgs + sessionIdArgument + processIdArgument)
                 {
-                    UseShellExecute = false,
+                    UseShellExecute = true,
```
timbussmann: I played around with those values again, I noticed the following:
* it doesn't seem to make a difference when running on the full framework.
* it doesn't seem to make a difference when running on .net core on a linux box
* when running the .net core tests via R#, it doesn't show the endpoint console when this is set to false.

based on that, I'd propose to stick with `true` for now unless there is some other problem caused by this?

src/PerformanceTests/Utils/Permutations/PermutationParser.cs
```diff
@@ -10,10 +10,10 @@ public static class PermutationParser
...
         public static string ToArgs(Permutation instance)
         {
-            return "--tests:" + string.Join(Separator, instance.Tests) + 
-                   " --category:" + instance.Category +
-                   " --description:\"" + instance.Description + "\"" +
-                   " --variables:" + ToString(instance);
+            return $"--tests:\"{string.Join(Separator, instance.Tests)}\"" +
+                   $" --category:\"{instance.Category}\"" +
+                   $" --description:\"{instance.Description}\"" +
+                   $" --variables:\"{ToString(instance)}\"";
```
timbussmann: linux can't deal with `;` separated values unless they are wrapped by "

src/PerformanceTests/Tests/Categories/Base.cs
ramonsmits: I'm fine with doing a code only approach for most values, especially these two. They could just be static readonly.

ramonsmits: Well, `InvokeEnabled` was used by the buildserver. Just make it a code var. Build would need to be addressed anyway as it is currently broken.

timbussmann:
not sure removing the config manager code makes this much easier since it's working now. Happy to remove it though ;)

src/PerformanceTests/Tests/Categories/Base.cs
```diff
@@ -92,23 +101,27 @@ static void LaunchAndWait(Permutation permutation, TestDescriptor testDescriptor
             var permutationArgs = PermutationParser.ToArgs(permutation);
             var sessionIdArgument = $" --sessionId={SessionId}";
 
+            //var useShellExecute = Environment.OSVersion.Platform != PlatformID.Win32NT;
```
danielmarbach: Should this be 🔥 ?

src/PerformanceTests/Tests/FodyWeavers.xml
```diff
@@ -1,4 +0,0 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
-<Weavers>
-  <Caseless />
```
danielmarbach: Why was this used and how do we know we can remove this now?

danielmarbach:
Made two comments but they can be addressed also after the PR is merged

src/PerformanceTests/Tests/Categories/Base.cs
timbussmann: yes, will do

src/PerformanceTests/Tests/FodyWeavers.xml
timbussmann: I haven't seen any explicit need for this. I checked the code for string comparisons using `==` and I found one potentially "vulnerable" case which I addressed in https://github.com/Particular/EndToEnd/pull/308

PR #308, opened 2017-12-05
Additional test changes

timbussmann:
some more changes required to run the tests on a linux box.

### DIFF
--- a/src/PerformanceTests/ArtifactsParser/Program.cs
+++ b/src/PerformanceTests/ArtifactsParser/Program.cs
@@ -23,7 +23,15 @@
             File.WriteAllText(dst, csv);
             Console.WriteLine("Parsed '{0}' recursively and written to '{1}'", src, dst);
 
-            Process.Start(dst);
+            try
+            {
+                // try to open the file
+                Process.Start(dst);
+            }
+            catch (Exception e)
+            {
+                // ignore exception
+            }
         }
     }
 }
--- a/src/PerformanceTests/ArtifactsParser/Program.cs
+++ b/src/PerformanceTests/ArtifactsParser/Program.cs
@@ -28,7 +28,7 @@
                 // try to open the file
                 Process.Start(dst);
             }
-            catch (Exception e)
+            catch (Exception)
             {
                 // ignore exception
             }
--- a/src/PerformanceTests/ArtifactsParser/Program.cs
+++ b/src/PerformanceTests/ArtifactsParser/Program.cs
@@ -23,14 +23,10 @@
             File.WriteAllText(dst, csv);
             Console.WriteLine("Parsed '{0}' recursively and written to '{1}'", src, dst);
 
-            try
+            if (Environment.OSVersion.Platform == PlatformID.Win32NT)
             {
                 // try to open the file
                 Process.Start(dst);
-            }
-            catch (Exception)
-            {
-                // ignore exception
             }
         }
     }
--- a/src/PerformanceTests/Persistence.V7.SQL.V3/Persistence.V7.Sql_v3.csproj
+++ b/src/PerformanceTests/Persistence.V7.SQL.V3/Persistence.V7.Sql_v3.csproj
@@ -14,8 +14,8 @@
 
   <ItemGroup>
     <PackageReference Include="NServiceBus" Version="7.0.0-*" />
-    <PackageReference Include="NServiceBus.Persistence.Sql" Version="3.0.0-*" />
-    <PackageReference Include="NServiceBus.Persistence.Sql.MsBuild" Version="3.0.0-*" />
+    <PackageReference Include="NServiceBus.Persistence.Sql" Version="4.0.0-*" />
+    <PackageReference Include="NServiceBus.Persistence.Sql.MsBuild" Version="4.0.0-*" />
   </ItemGroup>
 
   <ItemGroup>
--- a/src/PerformanceTests/Tests/Categories/AmazonSQSFixture.cs
+++ b/src/PerformanceTests/Tests/Categories/AmazonSQSFixture.cs
@@ -5,7 +5,7 @@ namespace Categories
     using Tests.Permutations;
     using Variables;
 
-    [TestFixture(Description = "Transports", Category = "Performance")]
+    [TestFixture(Description = "Transports", Category = "Performance"), Explicit]
     public class AmazonSQSFixture : Base
     {
         [TestCaseSource(nameof(CreatePermutations))]
--- a/src/PerformanceTests/Tests/Categories/AmazonSQSFixture.cs
+++ b/src/PerformanceTests/Tests/Categories/AmazonSQSFixture.cs
@@ -5,7 +5,7 @@ namespace Categories
     using Tests.Permutations;
     using Variables;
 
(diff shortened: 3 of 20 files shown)


src/PerformanceTests/Tests/Categories/AmazonSQSFixture.cs
```diff
@@ -5,7 +5,7 @@ namespace Categories
     using Tests.Permutations;
     using Variables;
 
-    [TestFixture(Description = "Transports", Category = "Performance"), Explicit]
+    [TestFixture(Description = "Transports", Category = "Performance")]
```
timbussmann: I removed the explicit attributes as the test adapter would ignore those tests otherwise, which doesn't seem to be what we want.

src/PerformanceTests/Tests/Categories/Base.cs
```diff
@@ -76,6 +76,7 @@ protected void Tasks(Permutation permutation, [CallerMemberName] string memberNa
         {
             var fixtureType = GetType();
             var fixture = fixtureType.GetCustomAttribute<TestFixtureAttribute>();
+            TestContext.WriteLine($"Running test {fixtureType} ({TestContext.CurrentContext.Test.FullName}/{TestContext.CurrentContext.Test.MethodName}");
```
timbussmann: that has turned out to be a very useful information when investigating failing tests as `dotnet test` only shows the failing test method, which can't easily be aligned with the failing test fixture.

src/PerformanceTests/Tests/Categories/AmazonSQSFixture.cs
ramonsmits: The transportsfixture, persistencefixture and transactionmodes fixture are the tests that need to run automated. All explicit onces are for testing purposes so that developers do not need to hunt for their specific permutations.

src/PerformanceTests/ArtifactsParser/Program.cs
```diff
@@ -23,7 +23,15 @@ static void Main(string[] args)
...
 
-            Process.Start(dst);
+            try
+            {
+                // try to open the file
+                Process.Start(dst);
+            }
+            catch (Exception e)
+            {
+                // ignore exception
```
ramonsmits: Can you log the exception so that atleast we see what happened?

src/PerformanceTests/Persistence.V7.SQL.V3/Persistence.V7.Sql_v3.csproj
```diff
@@ -14,8 +14,8 @@
 
   <ItemGroup>
     <PackageReference Include="NServiceBus" Version="7.0.0-*" />
-    <PackageReference Include="NServiceBus.Persistence.Sql" Version="3.0.0-*" />
```
ramonsmits: OMG... I already forgot that major version conflict... *sigh*

src/PerformanceTests/Tests/Categories/AmazonSQSFixture.cs
ramonsmits: See previous comment

src/PerformanceTests/Tests/Categories/ConcurrencyFixture.cs
```diff
@@ -6,7 +6,7 @@ namespace Categories
     using Tests.Permutations;
     using Variables;
 
-    [TestFixture(Description = "Concurrency", Category = "Performance"), Explicit]
+    [TestFixture(Description = "Concurrency", Category = "Performance")]
```
ramonsmits: Potentially we can run this... but these are ALOT of permutations but very valuable non the less. Only thing is that it takes really long before the results are in.

src/PerformanceTests/Tests/Categories/DistributorFixture.cs
```diff
@@ -6,7 +6,6 @@ namespace Categories
     using Variables;
 
     [TestFixture(Description = "Distributor distribution", Category = "Performance")]
-    [Explicit]
```
ramonsmits: Please add this again, this is purely for testing purposes.

src/PerformanceTests/Tests/Categories/SerializerFixture.cs
```diff
@@ -6,7 +6,7 @@
 namespace Categories
 {
 
-    [TestFixture(Description = "Serializer differences", Category = "Performance"), Explicit]
+    [TestFixture(Description = "Serializer differences", Category = "Performance")]
```
ramonsmits: If we want to run this for dotnet core then we must select a supported transport like RabbitMQ but because that is running remotely it is very hard to measure any difference in performance because serialization isn't the biggest issue.

src/PerformanceTests/Tests/Categories/SqlFixture.cs
```diff
@@ -5,7 +5,7 @@ namespace Categories
     using Tests.Permutations;
     using Variables;
 
-    [TestFixture(Category = "Performance"), Explicit]
+    [TestFixture(Category = "Performance")]
```
ramonsmits: Again, already covered by PersistenceFixture.

src/PerformanceTests/Tests/Categories/SqlTransportFixture.cs
```diff
@@ -5,7 +5,7 @@ namespace Categories
     using Tests.Permutations;
     using Variables;
 
-    [TestFixture(Description = "Transports", Category = "Performance"), Explicit]
+    [TestFixture(Description = "Transports", Category = "Performance")]
```
ramonsmits: Already covered by transport and transaction mode fixture

src/PerformanceTests/Tests/Categories/ConcurrencyFixture.cs
timbussmann: I'm happy to selectively exclude some tests (e.g. this one). So far I'm not sure how we could manually run those tests on linux though as the filter API on nunit is not documented.

src/PerformanceTests/Tests/Categories/DistributorFixture.cs
timbussmann: ok, will readd.

src/PerformanceTests/Tests/Categories/SqlFixture.cs
timbussmann: PersistenceFixture currently isn't configured for net core.

src/PerformanceTests/Tests/Categories/DistributorFixture.cs
timbussmann: although I'd rather question the point of these tests then tbh.

src/PerformanceTests/ArtifactsParser/Program.cs
danielmarbach: Let's if def then

timbussmann: changed to check the platform now. Not sure what will happen on the test windows VM which doesn't have excel installed though.

timbussmann:
I rolled back all `Explicit` attribute related changes to handle them in a dedicated PR. Please have another look at the changes and merge at will @ramonsmits

PR #309, opened 2017-12-05
Test configuration

timbussmann:
the test configuration to be used for the performance tests targeting v7

### DIFF
--- a/src/PerformanceTests/ArtifactsParser/Program.cs
+++ b/src/PerformanceTests/ArtifactsParser/Program.cs
@@ -1,7 +1,6 @@
 ﻿namespace ArtifactsParser
 {
     using System;
-    using System.Diagnostics;
     using System.IO;
     using System.Net;
 
@@ -22,12 +21,6 @@
             var dst = Path.Combine(src, $"perftest-report.{Environment.UserName}@{Dns.GetHostName()}.{DateTime.Now:yyyyMMddThhmmss}.tsv");
             File.WriteAllText(dst, csv);
             Console.WriteLine("Parsed '{0}' recursively and written to '{1}'", src, dst);
-
-            if (Environment.OSVersion.Platform == PlatformID.Win32NT)
-            {
-                // try to open the file
-                Process.Start(dst);
-            }
         }
     }
 }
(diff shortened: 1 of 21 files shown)


src/PerformanceTests/ArtifactsParser/Program.cs
```diff
@@ -22,12 +21,6 @@ static void Main(string[] args)
             var dst = Path.Combine(src, $"perftest-report.{Environment.UserName}@{Dns.GetHostName()}.{DateTime.Now:yyyyMMddThhmmss}.tsv");
             File.WriteAllText(dst, csv);
             Console.WriteLine("Parsed '{0}' recursively and written to '{1}'", src, dst);
-
-            if (Environment.OSVersion.Platform == PlatformID.Win32NT)
```
timbussmann: this code even threw on my local machine, so I removed it entirely

src/PerformanceTests/Utils/Permutations/Permutations.cs
```diff
@@ -7,10 +7,10 @@ public class Permutations
     {
         public Audit[] AuditModes = { Audit.Off };
         public MessageSize[] MessageSizes = { MessageSize.Tiny };
-        public NServiceBusVersion[] Versions = (NServiceBusVersion[])Enum.GetValues(typeof(NServiceBusVersion));
+        public NServiceBusVersion[] Versions = { NServiceBusVersion.V7 };
         public Outbox[] OutboxModes;
         public Persistence[] Persisters = { Persistence.InMemory };
-        public Platform[] Platforms = { Platform.NetFramework };
+        public Platform[] Platforms = { Platform.NetCore };
```
danielmarbach: Should we really lock in here to NetCore only?

src/PerformanceTests/Utils/Permutations/Permutations.cs
```diff
@@ -7,10 +7,10 @@ public class Permutations
     {
         public Audit[] AuditModes = { Audit.Off };
         public MessageSize[] MessageSizes = { MessageSize.Tiny };
-        public NServiceBusVersion[] Versions = (NServiceBusVersion[])Enum.GetValues(typeof(NServiceBusVersion));
+        public NServiceBusVersion[] Versions = { NServiceBusVersion.V7 };
```
danielmarbach: Should we really lock into V7 only?

PR #315, opened 2018-01-04
Get the tests to run

SzymonPobiega:
* Changed the target to `net452` since the main test project targets `net452`
 * Hard-coded the assembly path convention to look for `net452`. This and the above needs to be solved properly when we want to be able to run on .NET core
 * Minor fixes to the abstraction types
 * Removed (for now) version 4 from the list
 * Hard-coded (for now) the region to `us-east-1`


timbussmann:
> Removed (for now) version 4 from the list

Where did you remove v4 from the list?

timbussmann:
fixed the callback tests by adding the missing subscription tracking mechanism. Tests are passing now.

PR #318, opened 2020-03-31
GitHubSync update

particularbot:
This is an automated synchronization PR.

The following source template repositories were used:
* particular/RepoStandards


bording:
@andreasohlund @ramonsmits Not sure RepoSync should be enabled on this repo since it's using the old project system in places and hasn't generally been updated.

PR #321, opened 2020-09-22
add Particular.CodeRules

adamralph:


kbaley (approved):
This was a bit of a slog but it all looks okay to me. I'm assuming all the log4net redirects are because one of the packages in CodeRules relies on it?

adamralph:
> I'm assuming all the log4net redirects are because one of the packages in CodeRules relies on it?

No, it's just VS adding missing binding redirects while updating the `csproj` with the addition of CodeRules. It's unrelated to the changes we're making. Such is the beauty of the old `csproj`.

